PR #34, opened 2019-02-16
Fix mistakes in translation (JP)

wakiyamap:
Sorry.
This commit https://github.com/Blockstream/esplora/pull/30 , it was not fixed.

I fixed it.


shesek:
Thanks again!

greenaddress:
@wakiyamap thank you! also updated the live instance

wakiyamap:
thanks! 
I think that it is an easy-to-use block explorer.
I will advertise to people using Japanese.

PR #46, opened 2019-03-04
Dev

matthi01:
Updating UI for new recent transactions listing


shesek:
Added with separate commits in https://github.com/Blockstream/esplora/commit/e9bf9c192998e5ae577d661486100350bfe25bfc & https://github.com/Blockstream/esplora/commit/80c785aa8440b18938bd19ebe91fb517e4a55942.

PR #96, opened 2019-06-05
Logging files with rotation and also print to stdout/stderr

RCasatta:
At the moment processes launched by runit are printing to stdout and output can be seen from the docker image output.
This MR allows to have rotating logs file in /data/logs for each process (using standard runit tool svlogd) and also see log output from the docker image ouput for optional further processing.


greenaddress:
ack 77f505d12690df9099386348f9ce27982f50148e

greenaddress:
merged rebased

Issue #87, opened 2019-05-20
fee-estimate api returns confirmation target as string and not numeric as api doc suggests


wintercooled:
This: https://github.com/Blockstream/esplora/blob/master/API.md#fee-estimates

Says that the results of /api/fee-estimates is of the format:

{ 2: 36.183, 3: 34.841, 4: 34.841, 6: 34.841, 10: 22.164, 20: 9.692, 144: 1, 501: 1, 1008:1 }

When the actual output is: 

{"144":1.0,"1008":1.0,"4":32.934,"2":62.769,"10":24.383,"20":1.077,"3":59.354,"504":1.0,"6":25.608}

It would be better to have it as numeric as the documentation suggests as the results could be ordered easier.

wintercooled:
Raised Pull Request #97

shesek:
Closed in favor of #97

PR #97, opened 2019-06-06
Amend example results of fee-estimates to string: decimal

wintercooled:
The api itself returns string:decimal pairs but the documentation example shows it as int:decimal.

This brings the documentation in line with the api.

Addresses Issue #87

### DIFF
--- a/API.md
+++ b/API.md
@@ -191,7 +191,7 @@ and the value is the estimated feerate (in sat/vB).
 
 The available confirmation targets are 2, 3, 4, 6, 10, 20, 144, 504 and 1008 blocks.
 
-For example: `{ 2: 36.183, 3: 34.841, 4: 34.841, 6: 34.841, 10: 22.164, 20: 9.692, 144: 1, 501: 1, 1008:1 }`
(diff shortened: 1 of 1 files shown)


shesek:
Thanks! JSON object keys are always strings, so its indeed better to display it that way.

Two small nits, though: could you retain their ordering (lowest target first) and the spacing (after `,` and inside the `{}`)?

wintercooled:
The things is that because they are strings the example I added is how they are ordered in the results when you call the API. This is in fact the reason I raised the original issue as I was expecting ordering based on integers but because they are strings they don't return in that order.

shesek:
JSON objects have undefined ordering and in general should not be relied upon to have any specific order. The ordering you're currently seeing returned from the API is not guaranteed and could change, too. Since there's no real "right" order to use here, I prefer to optimize it for readability instead :)

wintercooled:
Understood - I have fixed and ordered then now. Thanks!

shesek:
Thanks @wintercooled! Merged.

(I accidentally hit the "revert" button, which created the revert commit above, but it was not merged into master)

PR #172, opened 2020-02-06
Fixed electrum port typo

bavarianledger:
changed from 500001 to 50001


greenaddress:
@bavarianledger thank you for catching this!

PR #224, opened 2020-07-06
Unclear how to access the API from reading this

matthiasdebernardini:
It was not clear to me how to access the API in the terminal from reading this document, therefore would like to add a minimal example of how to use esplora and curl together.

### DIFF
--- a/API.md
+++ b/API.md
@@ -2,6 +2,12 @@
 
 JSON over RESTful HTTP. Amounts are always represented in satoshis.
 
+To access the API via curl in the terminal insert `api` into the route after testnet or for mainnet after the domain name.
+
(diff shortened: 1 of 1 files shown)


matthiasdebernardini:
If this is too long, I think some mention of which route to take to access the API should be added

greenaddress:
Try to open this in a browser https://blockstream.info/api/blocks/tip/hash (or curl or wget or python urlib or python requests, etc)

greenaddress:
utack 9c633ec61e10230997538ccdca6f44bfb0a48663

shesek:
I added some more information about the public APIs in https://github.com/Blockstream/esplora/commit/0af3acab6267df8d1d3f74741044c05ad7ee3311.

PR #233, opened 2020-07-28
Light Theme - Assets table row text color update

adebayooriyomi:
On - https://blockstream.info/liquid/assets , switch to light mode.
**Current behavior**
The text displays in white. 
**Desired behavior** Match the color with the main block list (black).

See issue created in the link Below:
https://gl.blockstream.com/greenaddress/esplora/-/issues/96

### DIFF
--- a/client/src/views/block.js
+++ b/client/src/views/block.js
@@ -17,7 +17,7 @@ export default ({ t, block: b, blockStatus: status, blockTxs, openTx, spends, op
       <div className="container">
         { search({ t, klass: 'page-search-bar' }) }
         <div>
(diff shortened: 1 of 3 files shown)


shesek:
The table is switching to a striped design on small screen sizes (odd rows with light background, even ones with dark). This change looks well on the light backgronud rows, but not on the dark background ones:

image

Adding something like this (in addition to your current CSS) should do the trick:

```
@media only screen and (max-width: 850px) {
  .theme-light .assets-table-link-row:nth-child(odd)  .assets-table-row {
     color: #fff !important; 
   }
}
```

PR #261, opened 2020-10-06
Improved numeric spacing

adebayooriyomi:
**Improved numeric spacing**

Issue created by Allen
https://gl.blockstream.com/greenaddress/esplora/-/issues/99

- Remove Spacing and Commas from Main Block list
- Timestamp - relabel the header as the displayed timezone, e.g. TIMESTAMP (GMT -5), and each item should only display the date/time.

Transaction details:

* always display block height without a space or comma.
* amounts, always display 8 full decimals (unless 0).

### DIFF
--- a/client/src/views/home.js
+++ b/client/src/views/home.js
@@ -34,7 +34,7 @@ export const recentBlocks = ({ t, blocks, loading, ...S }) => homeLayout(
     <div className="blocks-table">
       <div className="blocks-table-row header">
         <div className="blocks-table-cell">{t`Height`}</div>
-        <div className="blocks-table-cell">{t`Timestamp`}</div>
+        <div className="blocks-table-cell">{process.browser ? t`Timestamp` : t`Timestamp (UTC)`}</div>
         <div className="blocks-table-cell">{t`Transactions`}</div>
         <div className="blocks-table-cell">{t`Size (KB)`}</div>
         <div className="blocks-table-cell">{t`Weight (KWU)`}</div>
@@ -42,8 +42,8 @@ export const recentBlocks = ({ t, blocks, loading, ...S }) => homeLayout(
       { blocks && blocks.map(b =>
         <div className="blocks-table-link-row">
         <a className="blocks-table-row block-data" href={`block/${b.id}`}>
-          <div className="blocks-table-cell highlighted-text" data-label={t`Height`}>{formatNumber(b.height)}</div>
-          <div className="blocks-table-cell" data-label={t`Timestamp`}>{formatTime(b.timestamp, t)}</div>
+          <div className="blocks-table-cell highlighted-text" data-label={t`Height`}>{b.height}</div>
+          <div className="blocks-table-cell" data-label={t`Timestamp`}>{formatTime(b.timestamp, false)}</div>
           <div className="blocks-table-cell" data-label={t`Transactions`}>{formatNumber(b.tx_count)}</div>
           <div className="blocks-table-cell" data-label={t`Size (KB)`}>{formatNumber(b.size/1000)}</div>
           <div className="blocks-table-cell" data-label={t`Weight (KWU)`}>{formatNumber(b.weight/1000)}</div>
--- a/client/src/views/util.js
+++ b/client/src/views/util.js
@@ -8,8 +8,8 @@ const DEFAULT_PRECISION = 0
 
 const pad = n => n < 10 ? '0'+n : n
 
-const formatTimezone = unix => {
-  const tzOffset = new Date(unix*1000).getTimezoneOffset() * -1;
+const formatTimezone = time => {
+  const tzOffset = time.getTimezoneOffset() * -1;
   return tzOffset == 0 ? 'UTC' : 'GMT ' + (tzOffset < 0 ? '' : '+') + (tzOffset/60)
 }
 
@@ -18,7 +18,7 @@ export const formatTime = (unix, with_tz = true) => {
 
   return `${time.getFullYear()}-${pad(time.getMonth() + 1)}-${pad(time.getDate())}`
        + ` ${pad(time.getHours())}:${pad(time.getMinutes())}:${pad(time.getSeconds())}`
-       + (with_tz ? ' ' +formatTimezone(unix) : '')
+       + (with_tz ? ' ' + formatTimezone(time) : '')
 }
 
 export const formatSat = (sats, label=nativeAssetLabel) => `${formatNumber(sat2btc(sats))} ${label}`
--- a/client/src/views/util.js
+++ b/client/src/views/util.js
@@ -8,13 +8,17 @@ const DEFAULT_PRECISION = 0
 
 const pad = n => n < 10 ? '0'+n : n
 
-export const formatTime = unix => {
+const formatTimezone = unix => {
+  const tzOffset = new Date(unix*1000).getTimezoneOffset() * -1;
+  return tzOffset == 0 ? 'UTC' : 'GMT ' + (tzOffset < 0 ? '' : '+') + (tzOffset/60)
+}
+
+export const formatTime = (unix, with_tz = true) => {
   const time = new Date(unix*1000)
-      , tzOffset = time.getTimezoneOffset() * -1
 
   return `${time.getFullYear()}-${pad(time.getMonth() + 1)}-${pad(time.getDate())}`
        + ` ${pad(time.getHours())}:${pad(time.getMinutes())}:${pad(time.getSeconds())}`
-       + ` GMT${tzOffset == 0 ? '' : (tzOffset < 0 ? '' : '+') + (tzOffset/60)}`
+       + (with_tz ? ' ' +formatTimezone(unix) : '')
 }
 
 export const formatSat = (sats, label=nativeAssetLabel) => `${formatNumber(sat2btc(sats))} ${label}`
--- a/client/src/views/block.js
+++ b/client/src/views/block.js
@@ -17,7 +17,7 @@ export default ({ t, block: b, blockStatus: status, blockTxs, openTx, spends, op
       <div className="container">
         { search({ t, klass: 'page-search-bar' }) }
         <div>
-          <h1 className="block-header-title">{t`Block ${formatNumber(b.height)}`}</h1>
+          <h1 className="block-header-title">{t`Block ${b.height}`}</h1>
           <div className="block-hash"><span>{b.id}</span>
             { process.browser && <div className="code-button">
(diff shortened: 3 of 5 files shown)


client/src/views/util.js
```diff
@@ -8,13 +8,29 @@ const DEFAULT_PRECISION = 0
...
+  
+  // Default Return Time and Timezone.... dataType === undefined. 
+  }else{
+      const time = new Date(unix*1000)
+        , tzOffset = time.getTimezoneOffset() * -1
+
+    return `${time.getFullYear()}-${pad(time.getMonth() + 1)}-${pad(time.getDate())}`
+        + ` ${pad(time.getHours())}:${pad(time.getMinutes())}:${pad(time.getSeconds())}`
+        + ` GMT${tzOffset == 0 ? '' : (tzOffset < 0 ? '' : '+') + (tzOffset/60)}`
+  }
```
shesek, line 33: I would go with something like this to reduce duplication, and to use a boolean flag instead of a dataType enum.

```js
export const formatTimezone = unix => {
  const tzOffset = new Date(unix*1000).getTimezoneOffset() * -1;
  return `GMT${tzOffset == 0 ? '' : (tzOffset < 0 ? '' : '+') + (tzOffset/60)}`
}

export const formatTime = (unix, with_tz) => {
  const time = new Date(unix*1000)

  return `${time.getFullYear()}-${pad(time.getMonth() + 1)}-${pad(time.getDate())}`
       + ` ${pad(time.getHours())}:${pad(time.getMinutes())}:${pad(time.getSeconds())}`
       + (with_tz ? ' ' +formatTimezone(unix) : '')
}
```

shesek:
> Remove Spacing and Commas from Main Block list

There a bunch of other places that display block heights - the pages for individual blocks, txs, addresses, assets, etc. Is the intention to avoid a thousands separator anywhere that a block height appears, or just specifically in some places?

> Timestamp - relabel the header as the displayed timezone, e.g. TIMESTAMP (GMT -5), and each item should only display the date/time.

Why? I'm not so sure that this is a good idea, using the timezone as the column's header would be confusing, at least for me. Is this a design pattern that you encountered anywhere else before?

> amounts, always display 8 full decimals (unless 0).

Note that the PR doesn't actually do that. (And also, why?)

--

Could you squash your commits so the `DS_Store` and `assets.json` files won't be part of the PR entirely, instead of checked-in and then removed?

adebayooriyomi:
Hi Shesek, I messaged Allen and Selene about your questions above and this was the response I got from Allen. See below.

**Allen**

> Shesek doesn't have access to gitlab. He may need the content of the request copied so he can understand what was requested.

_See Request Below._

**Improved numeric spacing**
--------------------------------------------------------------------------------
Numbers in Esplora could use some improvement.

Main Block list page:

- Height should not have any spaces or commas.
- Timestamp - relabel the header as the displayed timezone, e.g. TIMESTAMP (GMT -5), and each item should only display the date/time.

Block details:

- height should not have a space.
- amounts in each transaction - always display 8 full decimals (or 0, if the value is 0).

Transaction details:

- always display block height without a space or comma.
- amounts, always display 8 full decimals (unless 0).

Asset details:

- always display as locale dependent for amount issued. Note: circulating amount uses a different font than issued amount.
- as with transaction details, always use 8 full decimals (unless the asset itself has fewer decimals).

--------------------------------------------------------------------------------


----------------------------------------------------------------------------------
**RESPONSE to your previous comment from Allen**
----------------------------------------------------------------------------------


**Remove Spacing and Commas from Main Block list**


_Shesek_
> Is the intention to avoid a thousands separator anywhere that a block height appears, or just specifically in some places?

**Allen**

- The intent is to remove the spacing completely so the numbers can easily be copied/pasted. It may have had commas in the past. The change was made without any feedback from product, so I really don't care what the reasoning was.

-----------------------------------------------------------------------------

**Timestamp - relabel the header as the displayed timezone, e.g. TIMESTAMP (GMT -5), and each item should only display the date/time.**


_Shesek_
> Why? I'm not so sure that this is a good idea, using the timezone as the column's header would be confusing, at least for me. Is this a design pattern that you encountered anywhere else before?

**Allen**

- I'm also unconcerned about his feedback about the heading of columns. He should be reviewing the technical content of your changes (and ensuring they meet the requirements/request), not giving personal opinions.

-----------------------------------------------------------------------------

**amounts, always display 8 full decimals (unless 0).**


_Shesek_
> And also, why?


**Allen**

- The reason for always showing 8 decimals is it is very hard to know if something is 7 decimals or 8 just by looking, and you can easily mistake a case where the least significant digit is 0, and look smaller than it really is.

client/src/views/util.js
adebayooriyomi: Changes made. Thank you.

client/src/views/util.js
```diff
@@ -8,13 +8,17 @@ const DEFAULT_PRECISION = 0
 
 const pad = n => n < 10 ? '0'+n : n
 
-export const formatTime = unix => {
+export const formatTimezone = unix => {
+  const tzOffset = new Date(unix*1000).getTimezoneOffset() * -1;
+  return `GMT${tzOffset == 0 ? '' : (tzOffset < 0 ? '' : '+') + (tzOffset/60)}`
+}
+
+export const formatTime = (unix, with_tz) => {
```
shesek, line 16: `with_tz` should default to true, so that the other places that display timestamps include the timezone.

client/src/views/home.js
```diff
@@ -34,16 +34,16 @@ export const recentBlocks = ({ t, blocks, loading, ...S }) => homeLayout(
...
         <div className="blocks-table-cell">{t`Size (KB)`}</div>
         <div className="blocks-table-cell">{t`Weight (KWU)`}</div>
       </div>
       { blocks && blocks.map(b =>
         <div className="blocks-table-link-row">
         <a className="blocks-table-row block-data" href={`block/${b.id}`}>
-          <div className="blocks-table-cell highlighted-text" data-label={t`Height`}>{formatNumber(b.height)}</div>
-          <div className="blocks-table-cell" data-label={t`Timestamp`}>{formatTime(b.timestamp, t)}</div>
+          <div className="blocks-table-cell highlighted-text" data-label={t`Height`}>{b.height}</div>
+          <div className="blocks-table-cell" data-label={t`Timestamp`}>{formatTime(b.timestamp, "",t)}</div>
```
shesek, line 46: The `formatTime` invocation should be `formatTime(b.timestamp, false)`.

The `t` argument is a leftover from when timestamps formatting was locale-dependent and is no longer needed. There are two other places that still pass it, in `views/tx.js` and `views/block.js`, would you mind removing it from there too?

shesek (changes requested):

shesek:
Would you mind updating the commit message to also mention the changes made to the timezone display?

adebayooriyomi:
I have updated the commit message and removed the 't' from view.js and block.js.

client/src/views/home.js
adebayooriyomi: changes have been made. Thank you.

shesek:
Giving this some more thought, it seems that labeling the column with the timezone is problematic because of daylight savings, which means that the timezone offset is dependent on the timestamp itself and not just on the host machine's configurations. So for example, its possible to have two consecutive blocks where one is GMT+2 and the other is GMT+3, if daylight savings was changed right in between them.

client/src/views/home.js
```diff
@@ -34,16 +34,16 @@ export const recentBlocks = ({ t, blocks, loading, ...S }) => homeLayout(
     <div className="blocks-table">
       <div className="blocks-table-row header">
         <div className="blocks-table-cell">{t`Height`}</div>
-        <div className="blocks-table-cell">{t`Timestamp`}</div>
+        <div className="blocks-table-cell">{`${t`Timestamp`} (${formatTimezone(new Date().getTime())})`}</div>
```
shesek, line 37: `formatTimezone` expects a unix timestamp in seconds while `getTime()` returns it in milliseconds, so you should divide this by 1000. And you could also use the shorter `Date.now()`. So basically, `Date.now()/1000`.

But note that (as I mentioned on the [previous comment](https://github.com/Blockstream/esplora/pull/261#issuecomment-711434434)) using the timezone offset of the current timestamp is problematic because it might not match the timezone offset of the actual block timestamps.

shesek (changes requested):

client/src/views/home.js
adebayooriyomi: Changes have been made

client/src/views/home.js
```diff
@@ -1,7 +1,7 @@
 import Snabbdom from 'snabbdom-pragma'
 import layout from './layout'
 import search from './search'
-import { formatTime, formatSat, formatNumber } from './util'
+import { formatTime, formatTimezone, formatSat, formatNumber } from './util'
```
shesek, line 4: Importing `formatTimezone` is no longer necessary.

client/src/views/util.js
shesek: The default still needs to change here.

client/src/views/util.js
```diff
@@ -8,13 +8,17 @@ const DEFAULT_PRECISION = 0
 
 const pad = n => n < 10 ? '0'+n : n
 
-export const formatTime = unix => {
+export const formatTimezone = unix => {
```
shesek, line 11: I would consider removing the `export` now that `formatTimezone` is only used internally by `formatTime` and not by anything external.

shesek (changes requested):

client/src/views/util.js
adebayooriyomi: RE: The default still needs to change here. (Should I change the GMT in the formatTime to UTC)?

shesek: I meant changing the `with_tz` argument to default to true ([rational explained here](https://github.com/Blockstream/esplora/pull/261#discussion_r504283261)).

But yes, you could also change `formatTimezone` to return `UTC` instead of `GMT` when the timezone offset is 0, which would be more accurate.

adebayooriyomi: I added (with_tz = true) to the function's argument

shesek: I meant to use `UTC` only when the timezone offset is 0, i.e.

```js
return `${tzOffset == 0 ? 'UTC' : 'GMT ' + (tzOffset < 0 ? '' : '+') + (tzOffset/60)}`
```

adebayooriyomi: Updated.

shesek:
ACK 088b73c0ede3ac06f71b4977ceb6472731cea7e3

PR #273, opened 2020-11-10
doc: fix esplora's docker image name in README

harshjv:
Esplora's docker image is available at `blockstream/esplora`

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ docker build -t esplora .
 ```
 docker run -p 50001:50001 -p 8080:80 \
            --volume $PWD/data_bitcoin_mainnet:/data \
-           --rm -i -t esplora \
+           --rm -i -t blockstream/esplora \
            bash -c "/srv/explorer/run.sh bitcoin-mainnet explorer"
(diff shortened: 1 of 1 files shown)


shesek:
These `docker run` instructions refer to the [`docker build` instruction](https://github.com/Blockstream/esplora/blob/master/README.md#how-to-build-the-docker-image) above them, which would build this as an image named `esplora`.

Perhaps you could instead add a sentence under the *"How to build the Docker image"* section, saying something like "You can also use the pre-build image at `blockstream/esplora` instead of building it locally"?.

wj42420 (approved):

README.md
```diff
@@ -148,7 +148,7 @@ docker build -t esplora .
 ```
 docker run -p 50001:50001 -p 8080:80 \
```
wj42420, line 149: Good

wj42420 (changes requested):

philippem:
The docker image in the README refers to the image as built locally (using the instructions higher up in the README).

PR #278, opened 2020-11-20
Assets List Paging and Sorting

adebayooriyomi:
**Clean up Asset List Page**
**[Asset List] UI Changes - Nov 6**
- Remove search bar until Shesek make changes on server-side. (currently it can only search within the results on screen, which is confusing UX)
-  Replace "Load More" button to Pagination (serve as a stop-gap before we are able to implement a better way to search)
-  Show 50 results each page by default


client/src/app.js
```diff
@@ -51,7 +51,16 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
     , last_txids: parseHashes(loc.query.txids)
     , est_chain_seen_count: +loc.query.c || 0
     }))
-  , goAssetList$ = !process.env.IS_ELEMENTS || !process.env.ASSET_MAP_URL ? O.empty() : route('/assets')
+  //, goAssetList$ = !process.env.IS_ELEMENTS || !process.env.ASSET_MAP_URL ? O.empty() : route('/assets')
+  , goAssetList$ = !process.env.ASSETLIST_URL ? O.empty() : route(process.env.ASSETLIST_URL).map(loc => {
```
shesek, line 55: This doesn't need to use `process.env.ASSETLIST_URL`, you can check for `IS_ELEMENTS` and use a literal string for `route()`.

client/src/app.js
```diff
@@ -51,7 +51,16 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
...
+    return(
+        { startIndex: loc.query.start_index != null ? loc.query.start_index : 0,
```
shesek, line 57: This and `limit` below should be cast to a number and default to 0 if it isn't a valid number, with something like `loc.query.start_index != null && +loc.query.start_index || 0`, or even just `+loc.query.start_index || 0`

client/src/app.js
```diff
@@ -203,6 +212,12 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
         // use an empty object if the map fails loading for any reason
         .merge(extractErrors(HTTP.select('asset-map')).mapTo({}))
 
+  // Assets List State      
+  , assetList$ = !process.env.ASSETLIST_URL ? O.of([]) : reply('assetlist')
+  
+  // Get total Assets from API. If undefined, use 492 as default
+  , totalAssets$ = dropErrors(HTTP.select('assetlist')).map(resp => resp.headers['x-total-results'] || 492)
```
shesek, line 219: I would combine `assetList$` and `totalAssets$` into a single stream of `reply('assetlist').map(r => ({ assets: r.body, total: resp.headers['x-total-results'] }))`

client/src/app.js
```diff
@@ -241,7 +256,7 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
                      , mempool$, mempoolRecent$, feeEst$
                      , tx$, txAnalysis$, openTx$
                      , goAddr$, addr$, addrTxs$, addrQR$
-                     , assetMap$, goAsset$, asset$, assetTxs$
+                     , assetMap$, assetList$, goAssetList$, totalAssets$, goAsset$, asset$, assetTxs$
```
shesek, line 259: `goAssetList$` is an intent and does not belong in the state.

client/src/app.js
```diff
@@ -327,6 +342,12 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
     , !process.env.ASSET_MAP_URL ? O.empty() : O.of(
                                 { category: 'asset-map',  method: 'GET', path: process.env.ASSET_MAP_URL, bg: true })
 
+    //fetch asset List 
+    , goAssetList$.map(d => {
+      return ({ category: 'assetlist',  method: 'GET'
+          , path: `${process.env.ASSETLIST_URL}?sort_field=${d.sortField}&sort_dir=${d.sortDir}&limit=${d.limit}&start_index=${d.startIndex}` })
```
shesek, line 348: The string arguments (`sortField` and `sortDir`) need to be URL-escaped, which can be done with `encodeURIComponent()`.

Alternatively, you can pass the query string parameters as an object, which will automaticlaly take care of escaping it, and also be more readable. Like this

```
goAssetList$.map(d => ({ category: 'assetlist',  method: 'GET', path: 'assets/registry'
                       , query: { sort_field: d.sortField, sort_dir: d.sortDir, limit: d.limit, start_index: d.startIndex } }))
```

client/src/views/asset-list.js
```diff
@@ -1,12 +1,37 @@
...
+       let totalSupply = circulating == null ? t`Confidential`
+                    : formatAssetAmount(circulating, asset.precision, t) 
+    return totalSupply
+  }
 
-  const assets = Object.entries(assetMap)
-    .map(([ asset_id, [ domain, ticker, name ] ]) => ({ asset_id, domain, ticker, name  }))
-    .sort((a, b) => a.name.localeCompare(b.name))
+  const { startIndex, sortDir, sortField, limit } = goAssetList
+  const getSortDir = () => sortDir === 'asc' ? 'desc' : 'asc'
```
shesek, line 33: Why a function?

client/src/views/asset-list.js
```diff
@@ -1,12 +1,37 @@
...
+  const pageLink = `${process.env.ASSETLIST_URL}?sort_field=${sortField}&sort_dir=${sortDir}`
```
shesek, line 34: This should URL-encode the `sort_field` parameter with `encodeURIComponent()`.

`ASSETLIST_URL` can be replaced with the literal url.

client/src/views/asset-list.js
```diff
@@ -15,30 +40,116 @@ export default ({ assetMap, t, ...S }) => {
           <h1>{t`Registered assets`}</h1>
         </div>
       </div>
-
       <div className="container">
-        { !assets.length ? <p>{t`No registered assets`}</p>
-        : <div className="assets-table">
+        { !assetList ?  process.browser 
+            ? <div className="load-more-container"><img src="img/Loading.gif" /></div>
+            : <div className="load-more-container">
```
shesek, line 46: Why does this need to render an error here? If the request to load the assets fails, displaying an error should already be dealt with (via the `error$` stream).

I think this could be simplified to always show the loading indicator when the asset list isn't available, for both server and client side rendering (the server-side renderer waits for everything to load before capturing the html, so the loading indicator won't be visible).

client/src/views/footer.js
```diff
@@ -13,7 +13,7 @@ export default ({ t, page }) =>
       <div className="footer-logo">
         <div className="footer-links">
           { process.env.IS_ELEMENTS && [ <a href={`asset/${nativeAssetId}`}>Pegs</a>, ' | ' ] }
-          { process.env.ASSET_MAP_URL && [ <a href="assets">Assets</a>, ' | ' ] }
+          { process.env.ASSETLIST_URL && [ <a href={process.env.ASSETLIST_URL}>Assets</a>, ' | ' ] }
```
shesek, line 16: This can check for `IS_ELEMENTS` and use a literal url.

shesek (changes requested):

client/src/app.js
adebayooriyomi: I added it to state so that I can access the following params ( startIndex, sortDir, sortField, limit) 

I think it was easier that way...
const { startIndex, sortDir, sortField, limit } = goAssetList

client/src/views/asset-list.js
adebayooriyomi: I'm trying to write less code. 

I need this in three places (sortDir === 'asc' ? 'desc' : 'asc') So I made it a function instead.

No worries, I'll just copy and paste it 3 times.

client/src/app.js
adebayooriyomi: This will not work because reply return response.body

reply = (cat, raw) => dropErrors(HTTP.select(cat)).map(r => raw ? r : (r.body || r.text))

adebayooriyomi: Changes have been made...

client/src/app.js
adebayooriyomi: changes have been made

client/src/views/asset-list.js
adebayooriyomi: changes have been made

client/src/views/asset-list.js
adebayooriyomi: changes have been made

client/src/app.js
adebayooriyomi: changes have been made

client/src/app.js
```diff
@@ -203,6 +212,12 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
         // use an empty object if the map fails loading for any reason
         .merge(extractErrors(HTTP.select('asset-map')).mapTo({}))
 
+  // Assets List State      
+  , assetList$ = !process.env.ASSETLIST_URL ? O.of([]) : reply('assetlist')
```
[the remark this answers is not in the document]
adebayooriyomi: changes have been made

client/src/app.js
```diff
@@ -51,7 +51,16 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
     , last_txids: parseHashes(loc.query.txids)
     , est_chain_seen_count: +loc.query.c || 0
     }))
-  , goAssetList$ = !process.env.IS_ELEMENTS || !process.env.ASSET_MAP_URL ? O.empty() : route('/assets')
+  //, goAssetList$ = !process.env.IS_ELEMENTS || !process.env.ASSET_MAP_URL ? O.empty() : route('/assets')
+  , goAssetList$ = !process.env.ASSETLIST_URL ? O.empty() : route(process.env.ASSETLIST_URL).map(loc => {
+    return(
```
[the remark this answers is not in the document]
adebayooriyomi: changes have been made

client/src/app.js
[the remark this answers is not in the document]
adebayooriyomi: changes have been made

client/src/app.js
```diff
@@ -203,6 +209,10 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
         // use an empty object if the map fails loading for any reason
         .merge(extractErrors(HTTP.select('asset-map')).mapTo({}))
 
+  // Assets List State      
+  , assetList$ = !process.env.IS_ELEMENTS ? O.empty() : 
+      dropErrors(HTTP.select('assetlist')).map(r => ({ assets: r.body, query: r.request.query, total: r.headers['x-total-results'] || 492 }))
```
shesek, line 214: You can use `reply()` with the second argument set to true to receive the entire response and not just the body, i.e. `reply('assetlist', true).map(r => ...)`

client/src/views/asset-list.js
shesek: I meant why not a regular variable? For example `const reverseSortDir = sortDir === 'asc' ? 'desc' : 'asc'`

client/src/views/asset-list.js
```diff
@@ -1,12 +1,37 @@
...
+import { formatAssetAmount } from './util'
 import layout from './layout'
 
-export default ({ assetMap, t, ...S }) => {
+const staticRoot = process.env.STATIC_ROOT || ''
+export default ({ assetMap, assetList, loading, t, ...S }) => {
+
+  const { assets, query, total } = !assetList ? {} : assetList
+  const { start_index, sort_dir, sort_field, limit } = !query ? {} : query
+  const pageLink = `/assets/registry?sort_field=${encodeURIComponent(sort_field)}&sort_dir=${encodeURIComponent(sort_dir)}`
```
shesek, line 10: Link URLs should not begin with a leading `/` so that they'll be relative to the `<base href>` and not to the root domain.

This makes a difference when esplora is served from a subdirectory and not the `/` root path, like on the live website. If  `https://blockstream.info/liquid/` links to `/assets/registry`, it'll be relative to the root and lead to `https://blockstream.info/asset/registry/` without `liquid`.

client/src/views/asset-list.js
shesek: The url encoding looks good, added a separate comment with a followup.

client/src/views/footer.js
```diff
@@ -13,7 +13,7 @@ export default ({ t, page }) =>
       <div className="footer-logo">
         <div className="footer-links">
           { process.env.IS_ELEMENTS && [ <a href={`asset/${nativeAssetId}`}>Pegs</a>, ' | ' ] }
-          { process.env.ASSET_MAP_URL && [ <a href="assets">Assets</a>, ' | ' ] }
+          { process.env.IS_ELEMENTS && [ <a href="/assets/registry">Assets</a>, ' | ' ] }
```
shesek, line 16: Also no leading `/`.

client/src/views/asset-list.js
```diff
@@ -15,30 +40,109 @@ export default ({ assetMap, t, ...S }) => {
...
-        { !assets.length ? <p>{t`No registered assets`}</p>
-        : <div className="assets-table">
+        { !assets ? <div className="load-more-container"><img src="img/Loading.gif" /></div>
+          : <div className="assets-table">
               <div className="assets-table-row header">
-                <div className="assets-table-cell">{t`Name`}</div>
-                <div className="assets-table-cell ticker">{t`Ticker`}</div>
-                <div className="assets-table-cell">{t`Issuer domain`}</div>
-                <div className="assets-table-cell">{t`Asset ID`}</div>
+                <a href={`/assets/registry?sort_field=name&sort_dir=${encodeURIComponent(sort_dir === 'asc' ? 'desc' : 'asc')}`}
```
shesek, line 47: Also no leading `/` (here and in the two links below)

shesek (changes requested):

client/src/app.js
shesek: See https://github.com/Blockstream/esplora/pull/278#discussion_r530674311

client/src/views/asset-list.js
```diff
@@ -1,12 +1,37 @@
 import Snabbdom from 'snabbdom-pragma'
-
+import { formatAssetAmount } from './util'
 import layout from './layout'
 
-export default ({ assetMap, t, ...S }) => {
+const staticRoot = process.env.STATIC_ROOT || ''
+export default ({ assetMap, assetList, loading, t, ...S }) => {
```
shesek, line 6: `assetMap` isn't used here, if you don't unpack it to a variable here it'll remain part of `S` and then you won't have to specify it again below

client/src/app.js
shesek: Why are you keeping the request query on `assetList$` and later read it from the `asset-list` view, instead of reading it from `goAssetList`?

If that's because of my previous comment regarding putting `goAssetList` in the state - it was wrong, using `goAssetList` is fine. Please disregard what I said. :)

client/src/views/asset-list.js
```diff
@@ -1,12 +1,37 @@
...
+
+  const { assets, query, total } = !assetList ? {} : assetList
```
shesek, line 8: If the `assetList` isn't loaded yet, its better to avoid rendering the page entirely, rather than flashing the `No registered assets` message for a moment and then disappearing it when the assets finish loading.

A simple `if (!assetList) return;` should do the trick.

client/src/views/asset-list.js
```diff
@@ -1,12 +1,37 @@
...
+  const { start_index, sort_dir, sort_field, limit } = !query ? {} : query
+  const pageLink = `/assets/registry?sort_field=${encodeURIComponent(sort_field)}&sort_dir=${encodeURIComponent(sort_dir)}`
+  
+  const getSupply = (asset) => {
+    let { chain_stats = {}, mempool_stats = {} } = asset
+    let has_blinded_issuances =
+        chain_stats.has_blinded_issuances || mempool_stats.has_blinded_issuances
+    let is_native_asset = !asset.issuance_txin
+    let circulating = is_native_asset
```
shesek, line 17: Maybe worth extracting the circulating amount calculation to `util.js`?

shesek (changes requested):

client/src/views/asset-list.js
adebayooriyomi: changes have been made

client/src/views/asset-list.js
adebayooriyomi: When I added this line .... if (!assetList) return; The "loading spinner" wasn't showing.. 
When you navigate to the page the first time, it takes a few second for the AssetList to load. ..

const { assets, total } = !assetList ? {} : assetList
!assets ? <div className="load-more-container"></div>

client/src/views/asset-list.js
adebayooriyomi: changes have been made

client/src/views/footer.js
adebayooriyomi: changes have been made

client/src/views/asset-list.js
adebayooriyomi: changes have been made

client/src/app.js
adebayooriyomi: changes have been made.

client/src/app.js
```diff
@@ -51,7 +51,13 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
     , last_txids: parseHashes(loc.query.txids)
     , est_chain_seen_count: +loc.query.c || 0
     }))
-  , goAssetList$ = !process.env.IS_ELEMENTS || !process.env.ASSET_MAP_URL ? O.empty() : route('/assets')
+  //, goAssetList$ = !process.env.IS_ELEMENTS || !process.env.ASSET_MAP_URL ? O.empty() : route('/assets')
+  , goAssetList$ = !process.env.IS_ELEMENTS ? O.empty() : route('/assets/registry').map(loc => ({
```
shesek, line 55: Is it intended to be `/assets/registry` and not `/assets`?

client/src/views/asset-list.js
shesek: Yes, okay, makes sense. I see that you changed the code to not display the "No registered assets" message at all, but you could show it when `assets` is loaded but empty.

shesek (commented):
Looks good! Just a couple small things and its ready

client/src/views/asset-list.js
shesek: Yes, okay, makes sense. I see that you changed the code to not display the "No registered assets" message at all, but you could show it when `assets` is loaded and is an empty list.

client/src/app.js
adebayooriyomi: What do you mean?

client/src/views/asset-list.js
adebayooriyomi: changes have been made

client/src/app.js
shesek: The current asset list on the live website is at `https://blockstream.info/liquid/assets`, not `https://blockstream.info/liquid/assets/registry`. Did you mean to change it?

(the API endpoint does include `/registry`, but the frontend URL doesn't have to)

adebayooriyomi: I see... changes have been made

shesek:
ACK :+1:

client/src/app.js
piqjen-1xofFi-sikxyh: 😇

flavors/liquid-mainnet/extras.css
```diff
@@ -85,7 +85,7 @@
 .sortable{
   cursor: pointer;
 }
-.sortable::after{
+.desc::after{
```
shesek, line 88: What is `.desc`? I can't seem to find this class being used anywhere in the html

adebayooriyomi: const reverseSortDir = encodeURIComponent(sort_dir === 'asc' ? 'desc' : 'asc')

**html :**  <a href={`assets?sort_field=name&sort_dir=${reverseSortDir}`} 
                  className={`assets-table-cell sortable ${sort_field === "name" ? sort_dir : ""}`}>{t`Name`}
                </a>

shesek: My bad -- I was looking at the wrong deployment so didn't see it.

shesek:
ACK f74a289d57c68c2db658b9002a7170d9be2018b6

shesek (approved):

piqjen-1xofFi-sikxyh (approved):

piqjen-1xofFi-sikxyh (approved):

PR #331, opened 2021-08-09
Move lang switcher footer

adebayooriyomi:
- Add "word-break" to Asset List table row
- Move Language Switch to Footer
- Adjust table row padding
- Reduce loading gif size


shesek:
tACK https://github.com/adebayooriyomi/esplora/commit/00f4b543b3d4f8e4081308c457dfb301cc36cb2a (with some minor spacing fixups that I pushed to your branch)

PR #333, opened 2021-08-13
Allows private bridges to use torv3

DeviaVir:
This allows private bridges to use torv3 files and connect over torv3.


nitramiz (commented):
LGTM

run.sh
```diff
@@ -183,7 +183,15 @@ fi
 TORRCFILE="/srv/explorer/source/contrib/${DAEMON}-${NETWORK}-${MODE}-torrc"
 if [ -f $TORRCFILE ]; then
     cp $TORRCFILE /etc/tor/torrc
-    if [ -f /data/torrc ]; then
+
+    # torv2 being phased out, prefer v3
```
shesek, line 187: It says 'prefer v3', but the default appears to be v2?

Or does `TOR_VERSION` get set by default to v3 somewhere else?

DeviaVir: It's likely we are the only users of this, but for backwards compatibility (since it requires code changes) I kept v2 as the standard. Once we have rolled v3 out everywhere we use this we should just remove v2 as it is deprecated.

shesek: What does the comment mean then?

DeviaVir: More of a recommendation for the reader/user: prefer to use v3 since v2 is being phased out

PR #345, opened 2021-10-18
202101-testnet

adebayooriyomi:
Update Liquid Testnet UI elements color


www/style.css
```diff
@@ -1101,6 +1101,220 @@ body > .container, .navbar > .container, > .container, .content-wrap > div > .co
 
 .address-qr-code { width: 182px; }
 
+/* START ASSET TABLE */
```
shesek, line 1104: The CSS code that's shared for liquid mainnet and testnet could be moved into a new `flavors/liquid/extras.css` file, which can then be loaded by setting multiple `CUSTOM_CSS` files in the liquid networks config files. For example:

```
export CUSTOM_CSS="$CUSTOM_CSS flavors/liquid/extras.css flavors/liquid-testnet/extras.css"
```

This way, the liquid CSS code won't be bundled in the non-liquid builds where it isn't needed.

shesek (changes requested):

PR #350, opened 2021-11-03
Switch Liquid icon back to Tear drop icon, change all menu-logo icons…

adebayooriyomi:
Switch Liquid icon back to Tear drop icon, change all menu-logo icons…


shesek:
The `favicon`s for all the networks were removed, was that intentional?

adebayooriyomi:
Yes, I don't think we need it.
 It wasn't working and also it was a temporary solution for users to differentiate between the networks.

shesek:
ACK https://github.com/Blockstream/esplora/pull/350/commits/53fb685ecb23a4b6333af05faf36d662cc86fc75

PR #351, opened 2021-11-09
Update elements to 0.21.0

philippem:
- pass --jsonrpc-import to electrs for liquid regtest to workaround magic number changing
- when in liquid/regtest mode, create a wallet (elements 21 does not load a wallet by default)

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -67,6 +67,9 @@ else
         NGINX_REWRITE='rewrite ^/liquidregtest(/.*)$ $1 break;'
         NGINX_REWRITE_NOJS='rewrite ^/liquidregtest(/.*)$ " /liquidregtest/nojs$1?" permanent'
         NGINX_NOSLASH_PATH="liquidregtest"
+
+	# The --jsonrpc-import works around electrs not yet being aware of the magic number change for liquid regtest in elements 21.
+        ELECTRS_ARGS="$ELECTRS_ARGS --jsonrpc-import"
     elif [ "${NETWORK}" == "testnet" ]; then
         ELECTRS_NETWORK="liquidtestnet"
         PARENT_NETWORK="--parent-network regtest"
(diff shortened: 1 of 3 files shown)


run.sh
```diff
@@ -67,6 +67,8 @@ else
         NGINX_REWRITE='rewrite ^/liquidregtest(/.*)$ $1 break;'
         NGINX_REWRITE_NOJS='rewrite ^/liquidregtest(/.*)$ " /liquidregtest/nojs$1?" permanent'
         NGINX_NOSLASH_PATH="liquidregtest"
+
+        ELECTRS_ARGS="$ELECTRS_ARGS --jsonrpc-import"
```
shesek, line 71: It would be nicer if the magic bytes were configurable so the `--jsonrpc-import` workaround wasn't necessary, but I think its fine for now (and it also has the side benefit of being faster for small regtest chains, compared to reading from the blk files).

I would only suggest to add a comment here explaining why `--jsonrpc-import` is used.

run.sh
```diff
@@ -265,6 +267,7 @@ if [ "${NETWORK}" == "regtest" ]; then
     else
         /srv/explorer/$DAEMON/bin/${DAEMON}d -conf=/data/.$DAEMON.conf -datadir=/data/$DAEMON -daemon
     fi
+    cli -rpcwait createwallet "" || true
```
shesek, line 270: Something like `cli -rpcwait loadwallet "" || cli createwallet ""` would be better, so that it won't silently ignore errors other than 'this wallet already exists'.

run.sh
philippem: comment added in amendment.

run.sh
```diff
@@ -265,6 +268,7 @@ if [ "${NETWORK}" == "regtest" ]; then
     else
         /srv/explorer/$DAEMON/bin/${DAEMON}d -conf=/data/.$DAEMON.conf -datadir=/data/$DAEMON -daemon
     fi
+    cli -rpcwait loadwallet default || cli createwallet default
```
philippem, line 271: This has side effect of reporting an error

```
Elements Core starting
error code: -18
error message:
Wallet file verification failed. Failed to load database path '/data/liquid/liquidregtest/wallets/default'. Path does not exist.
```

shesek: The error could be silenced with something like `cli -rpcwait loadwallet default 2>/dev/null || ...`

greenaddress: the error is only in regtest anyway right? it can probably just be ignored

philippem: Added echo of message in amended commit

shesek:
ACK 5eff51e4b2b8daca4399ec0d293436f3c117a95f

PR #352, opened 2021-11-10
build bitcoin core from sources, add support for savefeeestimate sync, update tor

greenaddress:
@shesek mind reviewing? thanks

### DIFF
--- a/Dockerfile.deps
+++ b/Dockerfile.deps
@@ -37,13 +37,13 @@ RUN git clone --quiet --depth 1 --branch v22.0 --single-branch --recursive https
  && git checkout a0988140b71485ad12c3c3a4a9573f7c21b1eff8 \
  && git apply /${CORE_PATCH} \
  && (cd depends \
- && make HOST=x86_64-pc-linux-gnu NO_QT=1 -j $(grep ^processor /proc/cpuinfo | wc -l)) \
+ && make HOST=x86_64-pc-linux-gnu NO_QT=1 -j $(nproc --all)) \
  && ./autogen.sh \
  && CONFIG_SITE=$PWD/depends/x86_64-pc-linux-gnu/share/config.site ./configure --prefix=/srv/explorer/bitcoin --disable-man --disable-zmq --disable-qt --disable-gui-tests --disable-bench \
                 --enable-experimental-asm --without-utils --enable-util-cli --without-libs --with-daemon --disable-maintainer-mode \
                 --disable-glibc-back-compat --disable-ccache --disable-dependency-tracking --disable-tests --with-gui=no \
- && make -j $(grep ^processor /proc/cpuinfo | wc -l) \
- && make install -j $(grep ^processor /proc/cpuinfo | wc -l) \
+ && make -j $(nproc --all) \
+ && make install -j $(nproc --all) \
  && strip /srv/explorer/bitcoin/bin/* \
  && rm -fr ${CORE_SRC} /${CORE_PATCH})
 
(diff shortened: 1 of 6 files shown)


run.sh
```diff
@@ -279,6 +278,7 @@ if [ -n "$SYNC_SOURCE" ]; then
   fi
   # then fetch a recent mempool.dat,
   curl -f -s -u sync:$SYNC_SECRET -o $DAEMON_DIR/mempool.dat $SYNC_SOURCE || true
```
shesek, line 280: `SYNC_SOURCE` is currently expected to contain the full URL of the mempool file (i.e. `https://blockstream.info/_sync/mempool`). It should be changed to the URL of the `_sync` subdirectory (without the filename), then have it download `$SYNC_SOURCE/mempool` and `$SYNC_SOURCE/fee_estimates` here.

greenaddress: thanks didn't catch that, will fix!

Dockerfile.deps
```diff
@@ -47,18 +69,17 @@ RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain 1.56.1
 RUN source /root/.cargo/env \
  && mkdir -p /srv/explorer/electrs{,_liquid} \
  && git clone --no-checkout https://github.com/blockstream/electrs.git \
- && cd electrs \
+ && (cd electrs \
  && git checkout 703c6a20d52b61a234a18812503bd754d448992a \
  && cp contrib/popular-scripts.txt /srv/explorer \
  && cargo install --root /srv/explorer/electrs_bitcoin --locked --path . --features electrum-discovery \
  && cargo install --root /srv/explorer/electrs_liquid --locked --path . --features electrum-discovery,liquid \
- && cd .. \
```
shesek, line 55: The `rm` for the `electrs` directory on the line below wouldn't work well without the `cd ..`.

One way to fix it is to close the `)` the second `cargo install` instead of after `strip`.

Dockerfile.deps
```diff
@@ -15,23 +15,45 @@ RUN sh -c '[ -n "$NO_LIQUID" ] && mkdir -p /wally/wally_dist || ( \
...
-RUN curl -sL -o bitcoin.tar.gz "https://bitcoincore.org/bin/bitcoin-core-${VERSION_BITCOINCORE}/bitcoin-${VERSION_BITCOINCORE}-x86_64-linux-gnu.tar.gz" \
- && echo "${SHA256SUM_BITCOINCORE}  bitcoin.tar.gz" | sha256sum --check \
- && tar xzf bitcoin.tar.gz -C /srv/explorer \
- && ln -s "/srv/explorer/bitcoin-${VERSION_BITCOINCORE}" /srv/explorer/bitcoin \
- && rm bitcoin.tar.gz
+# Build core from sources until PR https://github.com/bitcoin/bitcoin/pull/23387 is merged
+ENV CORE_PATCH=contrib/0001-add-support-to-save-fee-estimates-without-shutting-d.patch
+ENV CORE_SRC=/root/bitcoin
+COPY ${CORE_PATCH} /${CORE_PATCH}
+RUN git clone --quiet --depth 1 --branch v22.0 --single-branch --recursive https://github.com/bitcoin/bitcoin.git ${CORE_SRC} \
```
shesek, line 35: It would be faster to download the [tag tarball from github](https://github.com/bitcoin/bitcoin/archive/refs/tags/v22.0.tar.gz) instead of using `git clone`. But speed isn't critical here so its also fine to keep as is.

greenaddress: true but then you'd need to apply the patch with patch or something like that

shesek:
ACK 7757f88811b1f352a350575ebeaf5790d726fe99

nitramiz:
rebase on `87b66cef`

PR #358, opened 2022-01-10
Update settings for liquid-testnet

philippem:
- add consensus parameters to avoid failures (evbparams)
- add additional connection
- remove checkblockindex. Improves IBD time

See https://liquidtestnet.com/


shesek:
Similar changes made it in #355 (seems to be exactly the same, except that #355 also added `liquidcert.com:18886` as a node)

Issue #216, opened 2020-06-11
Error 500 for style.css and error 404 for /api/ calls


tecnique2:
Hi all,

I have just installed esplora on Ubuntu 18.04 (following exactly the description).

```
git clone https://github.com/Blockstream/esplora && cd esplora
npm install
export API_URL=http://localhost:3000/
npm run dev-server
```

The API_URL points to blockstream.info, because electr is still indexing (and will be for some days).

`export API_URL=https://blockstream.info/api/`

If I start the server with: `npm run dev-server` everything runs without errors and messages: `HTTP server running on :::5000` at the end.

If I point my browser to localhost:5000 the /style.css file is not loaded and the terminal shows the following error for it:
```
GET /style.css 500 29.048 ms - 1108
TypeError: Cannot read property 'readFile' of undefined
    at map (/home/xxx/esplora/dev-server.js:44:80)
    at Array.map (<anonymous>)
    at _callee2$ (/home/xxx/esplora/dev-server.js:44:64)
    at tryCatch (/home/xxx/esplora/node_modules/@babel/node/node_modules/regenerator-runtime/runtime.js:45:40)
    at Generator.invoke [as _invoke] (/home/xxx/esplora/node_modules/@babel/node/node_modules/regenerator-runtime/runtime.js:274:22)
    at Generator.prototype.(anonymous function) [as next] (/home/xxx/esplora/node_modules/@babel/node/node_modules/regenerator-runtime/runtime.js:97:21)
    at asyncGeneratorStep (/home/xxx/esplora/dev-server.js:37:103)
    at _next (/home/xxx/esplora/dev-server.js:39:194)
    at /home/xxx/esplora/dev-server.js:39:364
    at new Promise (<anonymous>)
```

And if I point my browser to the /api (f.e. http://localhost/5000/api/address/XXXX) I only get a 404 message.

I have googled all this for a while and tried some reinstalls.
But I can't figure out, what is wrong here.
So I would be thankful for some help!

tecnique2:
Solved the style.css problem.
After upgrading node.js from v8.10.0 (stable) to v14.4.0 (latest) the style.css file works fine (status 200).
Now I only have to find a solution for the /api 404 error.

shesek:
The web server at `http://localhost:5000/` serves the client-side UI only, API requests should go directly to `http://localhost:3000/` and not via `http://localhost:5000/api/`.

[...]

PR #359, opened 2022-01-11
chore: add nvm support

escapedcat:
Support for [nvm](https://github.com/nvm-sh/nvm#intro) helps people like me to avoid [issues like this](https://github.com/Blockstream/esplora/issues/216#issuecomment-642538558) on their first install.  
It will automatically use node v14 if you are using nvm.  
Everything works like before if you don't use nvm. It's fine.  

Minor format changes on the README due to prettier-plugin. I can revert these if wanted.


philippem:
nvm is always there in the environment where people run node, but documenting it is outside the scope of this README.

escapedcat:
This was adding nvm support, not just documenting it. But no worries. Feels kinda outdated by now :D

PR #361, opened 2022-01-18
Update electrs to latest tip

philippem:
- use jsonrpc-import for liquid-testnet to avoid chain binary incompatibilities

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -78,7 +78,8 @@ else
         NGINX_REWRITE_NOJS='rewrite ^/liquidtestnet(/.*)$ " /liquidtestnet/nojs$1?" permanent'
         NGINX_NOSLASH_PATH="liquidtestnet"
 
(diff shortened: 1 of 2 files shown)


run.sh
```diff
@@ -78,7 +78,7 @@ else
         NGINX_REWRITE_NOJS='rewrite ^/liquidtestnet(/.*)$ " /liquidtestnet/nojs$1?" permanent'
         NGINX_NOSLASH_PATH="liquidtestnet"
 
-        ELECTRS_ARGS="$ELECTRS_ARGS --asset-db-path /srv/liquid-assets-db"
+        ELECTRS_ARGS="$ELECTRS_ARGS --jsonrpc-import --asset-db-path /srv/liquid-assets-db"
```
shesek, line 81: Maybe add a comment here to explain why `--jsonrpc-import` is used?

shesek:
ACK https://github.com/Blockstream/esplora/pull/361/commits/7d9a260a0a07824ca03cd02859a318344ff418e1

PR #365, opened 2022-02-15
support disabling of regtest mining with env var NO_REGTEST_MINING

philippem:


shesek:
ACK https://github.com/Blockstream/esplora/pull/365/commits/0681738e333a2aa748f42c08719873717d95c07e

PR #383, opened 2022-05-25
blockstream site: add favicon.ico to webroot

al-x:
This is a backwards-compatibility measure to satisfy certain antiquated webcrawlers which indirectly penalize this website for not serving this file. A side-effect should be to lower the number of HTTP 404s from ancient browsers. The favicon.ico file contains only the 32x32 size since the 16x16 version looks unacceptably fuzzy.


itwantid:
.patch

DeviaVir (approved):

PR #404, opened 2022-08-12
Update loading icon

adebayooriyomi:
**Update loading icon.**

- Change loading gif to css loading component.

### DIFF
--- a/client/src/views/addr.js
+++ b/client/src/views/addr.js
@@ -5,7 +5,7 @@ import search from './search'
 import { txBox } from './tx'
 import { formatSat, formatNumber } from './util'
 import { addrTxsPerPage as perPage, maxMempoolTxs } from '../const'
-import load from '../components/loading'
+import loader from '../components/loading'
 
 const staticRoot = process.env.STATIC_ROOT || ''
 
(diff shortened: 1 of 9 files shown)


client/src/views/loading.js
```diff
@@ -1,11 +1,10 @@
 import Snabbdom from 'snabbdom-pragma'
 import layout from './layout'
 
+
 export default ({ t, ...S }) => layout(<div>
-  <div className="loading-delay">
```
shesek, line 5: The `loading-delay` class was being used to make the loading indicator visible only when loading takes longer than 0.3s, so that it doesn't show up when it loads very quickly. Do we want to preserve this behavior?

It can be done by adding these CSS properties to `.spinner`:

```css
visibility: hidden;
animation: 0.3s showDelayed;
animation-fill-mode: forwards;
```

client/src/views/addr.js
```diff
@@ -5,6 +5,7 @@ import search from './search'
 import { txBox } from './tx'
 import { formatSat, formatNumber } from './util'
 import { addrTxsPerPage as perPage, maxMempoolTxs } from '../const'
+import load from '../components/loading'
```
shesek, line 8: I would consider calling this `loader` or `loading` instead of `load`, to make its purpose clearer

shesek (commented):
Looks great overall, added a couple minor comments

Scutua (approved):

Scutua (approved):

PR #466, opened 2023-09-01
Disable Elements unblinding in the server-side rendering environment

shesek:
Attempting to access the URL hash fragment (`location.hash`) on the server-side was causing an error, because it is not available there.

Unblinding should only be done on the client-side for privacy reasons anyway, so this feature can be disabled entirely on the server-side.


philippem:
tested locally, works

philippem (approved):

satoshinakamoto007:
> tested locally, works

PR #518, opened 2024-08-15
New typography and styling

sprihozhiy:
Updated the Blockstream Explorer with new fonts


www/style.css
```diff
@@ -2357,8 +2322,6 @@ dl.mempool-histogram .bar:before {
 .sub-nav a{
```
Randy808, line 2322:
```suggestion
.sub-nav .active {
  color: #ffffff;
  position: relative;
}

.sub-nav a {
  margin-right: 50px;
  color: #ffffff;
}

.sub-nav .active::after {
    position: absolute;
    bottom: -10px;
    content: '';
    border-bottom: solid 2px #ffbb00;
    width: 100%;
    display: block;
}
```

www/style.css
```diff
@@ -2357,8 +2322,6 @@ dl.mempool-histogram .bar:before {
...
   margin-right: 50px;
   color: #ffffff;
-  font-weight: bold;
-  line-height: 2.5;
 }
 
 .sub-nav .active {
```
Randy808, line 2327:
```suggestion
.sub-nav .active {
  color: #ffffff;
  position: relative;
}

.sub-nav .active::after {
  position: absolute;
  bottom: -10px;
  content: '';
  border-bottom: solid 2px #ffbb00;
  width: 100%;
  display: block;
}
```

Randy808:
Everything else looks good!

www/style.css
sprihozhiy: Thank you @Randy808 for pointing out that issue!

shesek:
Looks great overall, I like the changes. :) Found a few small bugs:

1. On the light version when above 900px wide, there's a little glitch to the right of the logo (appears to happen because the background is repeating):

    image

2. On the light version when below 900px wide, the horizontal logo repeats within a square:

    image

3. On the homepage "Latest transactions" when above 995px wide (such that the table format is used), the txid column overflows into the value column next to it:
 
    image

This was all tested on Firefox.

I'm not a designer, but I'll offer one comment on the looks (that you're free to ignore) - I feel that the transaction details area uses too much bold text and would consider reducing some of it.

image

sprihozhiy:
> Looks great overall, I like the changes. :) Found a few small bugs:
> 
> 1. On the light version when above 900px wide, there's a little glitch to the right of the logo (appears to happen because the background is repeating):
>    image
> 2. On the light version when below 900px wide, the horizontal logo repeats within a square:
>    image
> 3. On the homepage "Latest transactions" when above 995px wide (such that the table format is used), the txid column overflows into the value column next to it:
>    image
> 
> This was all tested on Firefox.
> 
> I'm not a designer, but I'll offer one comment on the looks (that you're free to ignore) - I feel that the transaction details area uses too much bold text and would consider reducing some of it.
> 
> image

As I see it, the bugs are not related to the typography update, but I will fix them for sure.

Screenshot 2024-08-20 at 1 09 32 PM

And I agree that bold text in the details looks too much. I'll talk to the designers. Thank you!

sprihozhiy:
And I've noticed a missing padding in the container with Block title, I'll fix that as well

sprihozhiy:
@shesek I can't reproduce the third issue, tried on Firefox and Chrome
Screenshot 2024-08-20 at 1 39 31 PM
Screenshot 2024-08-20 at 1 38 18 PM

Randy808:
I looked into the overflow and noticed @shesek's page rendered a different font, and on closer inspection my fonts also weren't the correct ones either. It turns out `Inter` was never being imported and the font was falling back to `sans-serif` which can be manually configured to be a different font within a browser (I imagine @shesek's browser was configured to use a monospace-like default). To fix it, the `Inter` import link needs to be moved to the top of the temporary-cdn.css file (right now it's imported after the `@font-face`)

Ref:
Screenshot 2024-08-20 at 5 52 12 PM

sprihozhiy:
@Randy808 we noticed that issue after @nitramiz deployed to the staging. I am waiting when designers share the font files and we use them instead of the fonts served on the Adobe side (e.g. Rigid Square)

shesek:
> the bugs are not related to the typography update

Ah I see, indeed 1&2 are not (I wonder how that slipped in tho oO), but 3 does appear related

> It turns out `Inter` was never being imported

This probably happens because we have a strict [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) that prevents loading third-party resources, everything on the page must be served from `blockstream.info`.

The header is added with an nginx directive, which could explain the discrepancy between running the server locally without nginx (through `dev-server` I assume?) and on staging/production with ngninx.

www/temporary-cdn.css
```diff
@@ -0,0 +1,332 @@
...
+
+@font-face { font-family: "rigid-square"; src: url("https://use.typekit.net/af/bb5c38/00000000000000007735c433/30/l?primer=7fa3915bdafdf03041871920a205bef951d72bf64dd4c4460fb992e3ecc3a862&fvd=n7&v=3") format("woff2"), url("https://use.typekit.net/af/bb5c38/00000000000000007735c433/30/d?primer=7fa3915bdafdf03041871920a205bef951d72bf64dd4c4460fb992e3ecc3a862&fvd=n7&v=3") format("woff"), url("https://use.typekit.net/af/bb5c38/00000000000000007735c433/30/a?primer=7fa3915bdafdf03041871920a205bef951d72bf64dd4c4460fb992e3ecc3a862&fvd=n7&v=3") format("opentype"); font-display: auto; font-style: normal; font-weight: 700; font-stretch: normal; }
+
+@import url('https://fonts.googleapis.com/css2?family=Inter:wght@300;400;500;600&display=swap&subset=latin');
```
Randy808, line 8: Just to formalize my last comment, this line needs to be moved to the top of this file (before the '@font-face's). Even if this file is going to be replaced, it'll make the changes to review locally (so others can take a look and sign off before this file is removed)

Randy808: I think the font size needs to be smaller to prevent wrapping. 

This is what it looks like after this change has been applied and the the Inter font loaded:

Firefox


Chrome
Screenshot 2024-08-22 at 12 26 48 PM

sprihozhiy: fixed

www/style.css
```diff
@@ -1,15 +1,15 @@
 @import url("font/inconsolata/index.css");
+@import url("temporary-cdn.css");
```
shesek, line 2: This should be removed

Randy808 (approved):

sprihozhiy:
updated

sprihozhiy:
It seems like the design team reviewed the changes, and everything is good to be deployed.

mookey86 (approved):

philippem (approved):

RCasatta (approved):

Theghost256 (approved):

PR #528, opened 2024-09-18
updated to catch missing 404 for not found routes

cjrutherford:
cycles doesn't expose the raw request, and I've yet to find the actual place that builds the response that is parsed in the render callback.


Theghost256 (approved):

philippem:
addressed in https://github.com/Blockstream/esplora/pull/530

PR #530, opened 2024-09-20
prerender: Wait for the loading state to complete, fix HTTP error status codes

shesek:
Prior to this fix, `render()` preemptively considered the app rendering to be final, before completing the transition out of the loading state entirely.

This resulted in notFound pages not being identified as such, returning them with a 200 OK status code instead of a 404 Not Found.

### DIFF
--- a/client/src/run-server.js
+++ b/client/src/run-server.js
@@ -31,7 +31,9 @@ export default function render(pathname, args='', body, locals={}, cb) {
     cb(null, data || {
       html: lastHtml
     , title: lastState.title
-    , status: lastState.view == 'notFound' ? 404 : lastState.error ? 400 : 200
+    , status: lastState.view == 'notFound' ? 404
+            : lastState.view == 'error' ? lastState.error.status || 400
+            : 200
     })
   }
 
--- a/client/src/run-server.js
+++ b/client/src/run-server.js
@@ -46,7 +46,7 @@ export default function render(pathname, args='', body, locals={}, cb) {
 
     lastState = S
 
-    if (S.loading || !S.isReady) {
+    if (S.view == 'loading' || S.loading > 0) {
       if (!seenLoading) {
         seenLoading = true
         clearTimeout(timeout)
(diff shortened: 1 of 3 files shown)


Theghost256 (approved):

Theghost256 (approved):

cjrutherford (approved):
this change appears to be unhooking the implementation from the `S.isReady` property of the state in favor of checking if the view is loading while still checking the `S.loading` property which appears to be the count of active http requests. the original would read as "If  loading is truthy or isReady is false" where the new version would read as "if the view is 'loading' or loading is more than 0" 

otherwise the timeout won't be set (because processing is most likely complete already)

philippem:
when testing this (regtest docker, signet container), 400s are returned by the prerender server, not 404s. 
Here's an example of request from within a bitcoin regtest container, first requesting genesis block, then nonexistent block (changed last digit of block id from 6 to 7)
```
root@f6b23cb988d3:/srv/explorer# !8
curl -i http://127.0.0.1:80/regtest/nojs/block/0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206 | grep HTTP
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  5286  100  5286    0     0  HTTP/1.1 200 OK:--:-- --:--:-- --:--:--     0
 135k      0 --:--:-- --:--:-- --:--:--  135k
root@f6b23cb988d3:/srv/explorer# !9
curl -i http://127.0.0.1:80/regtest/nojs/block/0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2207 | grep HTTP
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  2830  100  2830    0     0  97586      0 --:--:-- --:--:-- --:--:-- 94333
HTTP/1.1 400 Bad Request

```

philippem (changes requested):
returns 400s, 404s expected

shesek:
I see, there are actually two different 'not found' cases here:

- One is when there are no route handlers for the requested path, which results in the `notFound` view. This is what was originally reported and was fixed by this PR to properly return 404.

- The other is when there's a matching route to handle the request, but an error is returned when querying the API. This results in the `error` view, which was still always returned with a 400.

  I pushed d597900f8bcb71fbd8f8c21450a669d0c23800c8 with a fix that retains the original API status code error in the `error$` stream, then uses it for the prerender server response. This applies to 404s but also to any other error code returned from the API.

philippem (approved):

PR #538, opened 2024-11-22
Docker: make available as ARM64

DeviaVir:
Relies on: https://github.com/ElementsProject/libwally-core/pull/465, for now, we have pushed to a "fork" on `blockstream/`..

Test images available on `blockstream/esplora`.


Dockerfile.deps
```diff
@@ -1,6 +1,6 @@
 # Build libwally wasm files. Used for client-side blinding verification on Elements-based chains
 
-FROM greenaddress/wallycore@sha256:d63d222be12f6b2535e1548aa7f21cf649e2230d6c9e01bd518c23eb0bccd46f AS libwally-wasm
+FROM blockstream/wallycore@sha256:6a41958eefaa3920846e423885e807d41eb0f1ae54e28f039d62d5c0f97003f5 AS libwally-wasm
```
RCasatta, line 3: Should this now become 956b107d688f549c6e3884424991b7d3d34d84173990d43046fd760d7918db7c after https://github.com/ElementsProject/libwally-core/pull/467 ?

Dockerfile.deps
```diff
@@ -23,42 +23,37 @@ RUN mkdir -p /srv/explorer \
...
 
 
-RUN git clone --quiet --depth 1 --single-branch --branch v0.39.0 https://github.com/nvm-sh/nvm.git /root/.nvm \
+RUN git clone --quiet --depth 1 --single-branch --branch v0.40.1 https://github.com/nvm-sh/nvm.git /root/.nvm \
  && rm -rf /root/.nvm/.git \
  && source /root/.nvm/nvm.sh \
- && nvm install v17.1.0
+ && nvm install v18.20.4
 
-# Build core from sources until PR https://github.com/bitcoin/bitcoin/pull/23387 is merged
```
RCasatta, line 31: This doesn't seem merged, are we currently not using `savemempoolfees`?
or is it ok to upgrade to version `27.0` without this?

RCasatta: It seems it's used (or at least there are reference) in the codebase: https://github.com/Blockstream/esplora/blob/4f5c82fa2a1dd00a6b4b40804c4b44b3532469cb/contrib/nginx-sync.conf.in#L18

DeviaVir: We needed this "complexity" because we had many esplora nodes come and go, and so needed a quick way to sync the fee-estimates, that's the os.execute in nginx you're seeing here.

I think we no longer need it, we don't use it in the new architecture either.

Dockerfile.deps
DeviaVir: Yes, super new image. Still checking with Jon - looks like there now is _only_ an ARM image.

DeviaVir: One more followup: https://github.com/ElementsProject/libwally-core/pull/468
and noticed that they pushed either AMD or ARM, but we can't have an if/else here so still using a blockstream source for now.

Dockerfile.deps
shesek: The docker `run.sh` entrypoint is using the `fee_estimates` endpoint:

https://github.com/Blockstream/esplora/blob/4f5c82fa2a1dd00a6b4b40804c4b44b3532469cb/run.sh#L301

If we're removing this, it should be removed there too. Also, it would be nice if the `fee_estimates` removal was separated to its own commit (or PR, even).

> we don't use it in the new architecture either

You mean that `SYNC_{SOURCE,SECRET}` (`run.sh`'s 4th/5th arguments) are no longer set in the new architecture?

If that's the case then the mempool endpoint is no longer used either, should we removed it too?

/cc @greenaddress

philippem (dismissed):

Dockerfile.deps
DeviaVir: I wondered about leaving mempool in for others that may find it useful, since it doesn't force us to apply a patch and do a custom build.

DeviaVir: Updated the PR as requested.

PR #561, opened 2025-03-27
update overpaying field to detect liquid discount

cjrutherford:
* extracted txAnalysis combine latest final function
* extracted calculateOverpaying func


philippem:
implemented in https://github.com/Blockstream/esplora/pull/562

PR #562, opened 2025-03-28
extend overpayment warning to discountCT (liquid) txes

philippem:
- when calculating overpayment, consider liquid transactions for which discount CT rates were or are available

- extract calculations for fee rates and overpayment into functions

### DIFF
--- a/client/src/app.js
+++ b/client/src/app.js
@@ -5,7 +5,24 @@ import {setAdapt} from '@cycle/run/lib/adapt';
 import { getMempoolDepth, getConfEstimate, calcSegwitFeeGains } from './lib/fees'
 import getPrivacyAnalysis from './lib/privacy-analysis'
 import { nativeAssetId, blockTxsPerPage, blocksPerPage } from './const'
-import { dbg, combine, extractErrors, dropErrors, last, updateQuery, notNully, processGoAddr, parseHashes, isHash256, makeAddressQR, tickWhileFocused, tickWhileViewing, updateBlocks } from './util'
+import {
+    dbg,
+    combine,
+    extractErrors,
+    dropErrors,
+    last,
+    updateQuery,
+    notNully,
+    processGoAddr,
+    parseHashes,
+    isHash256,
+    makeAddressQR,
+    tickWhileFocused,
+    tickWhileViewing,
+    updateBlocks,
+    calculateFeerates,
+    calculateOverpayment,
+} from './util'
 import l10n, { defaultLang } from './l10n'
 import * as views from './views'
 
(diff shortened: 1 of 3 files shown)


Randy808 (approved):

client/src/util.js
```diff
@@ -60,6 +60,22 @@ export const outTotal = tx => tx.vout.reduce((N, vout) => N + (vout.value || 0),
 
 export const isNativeOut = vout => (!vout.asset && !vout.assetcommitment) || vout.asset === nativeAssetId
 
+export const calculateFeerates = (tx, mempool, feeEst) => {
+  const feerate = tx.fee ? tx.fee / tx.weight * 4 : null;
+  const discountFeerate = tx.fee & tx.discount_vsize ? tx.fee / tx.discount_vsize : null;
```
RCasatta, line 65: is that a bitwise operator?

delta1: Good catch 

```suggestion
  const discountFeerate = tx.fee && tx.discount_vsize ? tx.fee / tx.discount_vsize : null;
```

philippem: wow. thanks

philippem: fixed in amended commit.

client/src/app.js
```diff
@@ -191,15 +208,14 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
...
-      .combineLatest(mempool$, feeEst$, (tx, mempool, feeEst) =>
-          ({ tx, mempool, feeEst, feerate: tx.fee ? tx.fee / tx.weight * 4 : null }))
-      .map(({ tx, feerate, mempool, feeEst }) => ({
+      .combineLatest(mempool$, feeEst$, calculateFeerates)
+      .map(({ tx, feerate, mempool, feeEst, discountFeerate}) => ({
         feerate
       , privacyAnalysis: getPrivacyAnalysis(tx)
       , segwitGains: calcSegwitFeeGains(tx)
       , mempoolDepth: !tx.status.confirmed && feerate != null && mempool ? getMempoolDepth(mempool.fee_histogram, feerate) : null
       , confEstimate: !tx.status.confirmed && feerate != null && feeEst ? getConfEstimate(feeEst, feerate) : null
```
shesek, line 217: Shouldn't `mempoolDepth` and `confEstimate` use the `discountedFeerate` too?

It might be more convenient to implement this if `calculateFeerates` added an `effectiveFeerate` field that is either the `discountFeerate` when available or the regular `feerate` otherwise.

Then we can use it for `calculateOverpayment()`, `getMempoolDepth()` and `getConfEstimate()` instead of having each function deal with both (as `calculateOverpayment()` currently does).

philippem: I have followed this uggestion and added it into an additional commit. I will squash after review feedback.

client/src/app.js
```diff
@@ -191,15 +208,14 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
...
+      .map(({ tx, feerate, mempool, feeEst, effectiveFeerate}) => ({
         feerate
       , privacyAnalysis: getPrivacyAnalysis(tx)
       , segwitGains: calcSegwitFeeGains(tx)
-      , mempoolDepth: !tx.status.confirmed && feerate != null && mempool ? getMempoolDepth(mempool.fee_histogram, feerate) : null
-      , confEstimate: !tx.status.confirmed && feerate != null && feeEst ? getConfEstimate(feeEst, feerate) : null
-      , overpaying: !tx.status.confirmed && feerate != null && feeEst && feeEst[2] != null ? feerate/feeEst[2] : null
+      , mempoolDepth: !tx.status.confirmed && feerate != null && mempool ? getMempoolDepth(mempool.fee_histogram, effectiveFeerate) : null
+      , confEstimate: !tx.status.confirmed && feerate != null && feeEst ? getConfEstimate(feeEst, effectiveFeerate) : null
+      , overpaying: !tx.status.confirmed ? calculateOverpayment(effectiveFeerate, feeEst) : null
```
shesek, line 218: Seems this should have the `feerate != null` test too

philippem: thank you, applied.

Rebased to upstream, squashed into 1 commit.

shesek (approved):

RCasatta (approved):

PR #565, opened 2025-04-07
Update liquid mainnet and liquid testnet elements conf

philippem:


shesek:
> update elements to latest released version

I verified the shasums match the downloads available on `https://github.com/ElementsProject/elements/releases/tag/elements-23.2.7` and that they're signed by psgreco's key (`2F656B0610604482`).

> update liquid testnet elements.conf to minimum necessary

I don't know enough about Elements to verify these changes.

However I did notice that you changed from `listen=0` to `listen=1`, was that intentional?

philippem:
> I don't know enough about Elements to verify these changes.

We've confirmed them.

> 
> However I did notice that you changed from `listen=0` to `listen=1`, was that intentional?

Yes, this change is intentional to give the liquid chains behaviour that is consistent with the bitcoin mainnet configuration.
It should be noted that networking access to the daemon running inside the container would be necessary in order to allow p2p connections to other daemons.

psgreco (approved):

shesek (approved):
ACK 577906405d7b126eee5c97942645318c7632f41c

PR #594, opened 2025-09-22
fix(prerenderer): HEAD requests cause double requests

DeviaVir:
`HEAD /nojs/address/... took 2530ms and returned 878259 bytes`

This proves HEAD requests are generating the full HTML body (878KB) and taking 2.5 seconds - the same expensive operation as GET requests.

The code confirms this: in server.js, every request (HEAD or GET) calls render(), which runs the full Cycle.js app to generate HTML. Express handles HEAD by not sending the body, but the work is still done.

The proxy uses HEAD requests to check if pages exist (200 vs 404/500), not to get content-length. The prerender determines status codes without needing the full HTML body.


prerender-server/client
```diff
@@ -1 +1 @@
-../client/src
```
philippem, line 1: why this change?

DeviaVir: We're running against the source code instead of the compiled code now:

* The prerender-server imports from `../client/run-server`
* The client code is written in ES6+ and needs Babel transpilation
* build.sh runs Babel to transpile `client/src` → `client/dist`
* The symlink ensures the `prerender-server` uses the transpiled/built code, not the raw source

PR #601, opened 2025-10-22
rename L-BTC to LBTC

philippem:
- also test variants


philippem:
> You're missing an 'L' before the second 'BTC' in the PR title

fixed in amended commit

shesek (approved):
Looks good 👍

 I grepped for other occurrences of 'L-BTC' (case-insensitively) and couldn't find any.

PR #612, opened 2026-03-03
Change the year on footer to be dynamic

SusanGithaigaN:
### Problem
The footer currently displays a hardcoded year: `© 2025 Blockstream Corporation Inc. All rights reserved` (see screenshot below).

hardcoded-year

This means the year has to be manually updated every year in `flavors/blockstream/config.env`.

### Solution
I replaced the hardcoded year with a dynamic value that automatically updates on every page load.

dynamic-year-value


shesek:
This could be implemented more simply by using `date +%Y` in the `SITE_FOOTER` under the `config.env` file (which is `source`d and can contain bash scripting).

SusanGithaigaN:
> This could be implemented more simply by using `date +%Y` in the `SITE_FOOTER` under the `config.env` file (which is `source`d and can contain bash scripting).

@shesek, I've made the update

shesek:
Looks good, thank you!

Can you drop the whitespace change in `footer.js`? Then it will be ready to merge.

SusanGithaigaN:
> Looks good, thank you!
> 
> Can you drop the whitespace change in `footer.js`? Then it will be ready to merge.

Done

shesek (approved):

PR #615, opened 2026-04-08
update bitcoin core, elements, and electrs versions

philippem:


Randy808 (approved):

README.md
```diff
@@ -140,6 +140,14 @@ be configured to the URL reachable by the server, typically `http://localhost:30
 
 ## How to build the Docker image
 
+Build the base image first:
```
shesek, line 143: Makes sense to mention this here.

I think this could come instead of the [existing instructions that appear below](https://github.com/philippem/esplora/blob/ef1c87307a4034a0bccb861c0ac80c3d40755d4e/README.md#build-new-esplora-base)?

shesek:
Elements v23.3.3 was released a couple days ago, should we upgrade to it?

README.md
philippem: imo it's ok to duplicate

philippem: yes it does make sense to remove instructions that appear after because it describes pushing as well. will remove that in a later PR.

philippem:
> Elements v23.3.3 was released a couple days ago, should we upgrade to it?

yes will in a later PR

PR #617, opened 2026-04-09
Updated navbar to use latest design.

Randy808:
```
# Bitcoin
export PORT=4999
npm run dev-server
```

```
# Liquid
export PORT=5000
npm run dev-server
```

# Mainnet
Current design:
bitcoin (old)

New design:
bitcoin (new)

Liquid light-theme

# Testnet
Current design:
bitcoin testnet (old)

New design:
bitcoin testnet (new)


shesek:
Instead of adding the `BASE_HREF` to each route handler, we could put the handlers on an express [`Router`](https://expressjs.com/en/4x/api.html#router) and mount that under the `BASE_HREF`. Something like `router = express.Router()`, then change all `app.{get,use}()` to `router.{get,use}()`, and finally `app.use(process.env.BASE_HREF||'/', router)`.

Alternatively, we could mount a pre-processing middleware that strips the `BASE_HREF` prefix before forwarding the request to the other handlers. Something like this (untested):

```js
if (process.env.BASE_HREF && process.env.BASE_HREF != '/') {
  const baseHref = process.env.BASE_HREF.replace(/\/+$/, '')
  app.use((req, res, next) => {
    if (req.url == '/' || req.url == baseHref) {
      res.redirect(302, baseHref+'/')
    } else if (req.url.startsWith(baseHref+'/')) {
      req.url = req.url.slice(baseHref.length)
      next()
    } else {
      res.sendStatus(404)
    }
  })
}
```

Personally I've just been making sure to run the dev-server with `BASE_HREF=/`, with e.g. `(source config1.env && source config2.env && BASE_HREF=/ npm run dev-server)`. But I guess it could make sense to support `BASE_HREF`, to make the dev-server behave more closely to the live server.

Randy808:
Mounted router onto base path and removed prefix from individual routes

philippem:
lgtm

philippem (dismissed):

FedOken:
image
1. You might want to add some spacing between the input field and the clear icon; the text is overlapping the clear icon.
2. You could add `cursor: pointer` to the clear icon's hover state.
3. The placeholder no longer fits in the input field; it might be worth shortening it

FedOken:
image
1. This behavior might be expected, but when we click on the network, we get the URL `/search?q=liquid`, whereas the main site uses `/liquid`.
2. Also, when we on Liquid env, clicking on the mainnet does not return a “Not Found” page.

FedOken:
image
Here, I might have made the indentation a little deeper before the circle. I might also have removed it when we're on the assets page the text highlight at the bottom extends into the circle, which doesn't look quite right in this case.

FedOken:
image
It doesn't look right on a white background

FedOken:
image
image
1. Review the breakpoints
2. On mobile, clicking the menu icon refreshes the page instead of opening it
3. The sidebar close button doesn't work in tablet view; this is also an issue for the prod

FedOken:
Otherwise, it's good - good work

Randy808:
@FedOken 
- The clear icon comes from the default HTML input behavior rather than something custom. There isn’t currently spacing between the text and icon, so I can look into whether we can safely adjust that without overriding native styles too heavily.
- I can also check if it’s possible to add a hover cursor to the clear icon, though I’m aiming to avoid large overrides since this PR is focused on gradually introducing the new designs rather than doing broader cleanup.
- I’ll pass along the placeholder text feedback to design so we can keep that consistent with the intended spec.
- The dev server behaves a bit differently from production, so it doesn’t currently support switching themes or API URLs per instance.
- For the navbar spacing and circle issues, I’ve tried to keep changes minimal in this PR to make review easier. I did prevent the navbar text from wrapping, which can truncate "Explorer API" on smaller screens. I can look at adjusting the breakpoint address that more cleanly.

Randy808:
Updated PR to add proper stylings for light theme and raised the target width for nav link line wraps. I also fixed the issue of the mobile menu not closing on certain screen sizes since it only required a [small](https://github.com/Blockstream/esplora/pull/617/changes/5844633af09906e9f812ad0e5ed375fe1ec16c6f#diff-3f710416dbe04f0fa44aeaef9817a13d98f8ed47f63c3c1615adfdddcca8ec06L2520-L2541) change. 

EDIT: Also addressed @sprihozhiy brought up regarding the page refresh when the network selector was touched on mobile

FedOken:
> @FedOken
> 
> * The clear icon comes from the default HTML input behavior rather than something custom. There isn’t currently spacing between the text and icon, so I can look into whether we can safely adjust that without overriding native styles too heavily.
image

```
input[type="search"]::-webkit-search-cancel-button {
    cursor: pointer;
    margin-left: 12px;
}
```

> * I can also check if it’s possible to add a hover cursor to the clear icon, though I’m aiming to avoid large overrides since this PR is focused on gradually introducing the new designs rather than doing broader cleanup.

The solution for the pointer icon above. I understand that this is a redesign overall, but we can fix the bugs related to this part.

> * I’ll pass along the placeholder text feedback to design so we can keep that consistent with the intended spec.

Ok

> * The dev server behaves a bit differently from production, so it doesn’t currently support switching themes or API URLs per instance.

Ok

> * For the navbar spacing and **circle issues**, I’ve tried to keep changes minimal in this PR to make review easier. 

It’s up to you, but this seems a bit odd. Maybe it’s just me.

> * I did prevent the navbar text from wrapping, which can truncate "Explorer API" on smaller screens. I can look at adjusting the breakpoint address that more cleanly.

It looks better; you could also add a gap to avoid that
Before
image
After
image

> *Updated PR to add proper stylings for light theme and raised the target width for nav link line wraps. I also fixed the issue of the mobile menu not closing on certain screen sizes since it only required a small change.

Looks good, thanks

> *EDIT: Also addressed @sprihozhiy brought up regarding the page refresh when the network selector was touched on mobile

It's working now; I don't know if you changed anything

client/src/views/network-selection.js
```diff
@@ -0,0 +1,66 @@
...
+              width="17"
+              height="9"
+              viewBox="0 0 17 9"
+              fill="none"
+              xmlns="http://www.w3.org/2000/svg"
+            >
+              <path d="M0.219933 0.220005C0.289589 0.150273 0.372307 0.0949535 0.463356 0.05721C0.554403 0.0194674 0.651998 4.1008e-05 0.750559 4.1008e-05C0.84912 4.1008e-05 0.946717 0.0194674 1.03777 0.05721C1.12881 0.0949535 1.21153 0.150273 1.28118 0.220005L8.25056 7.19032L15.2199 0.220005C15.3607 0.0792751 15.5515 0.000212669 15.7506 0.000212669C15.9496 0.000212669 16.1405 0.0792751 16.2812 0.220005C16.4219 0.360736 16.501 0.551608 16.501 0.75063C16.501 0.949654 16.4219 1.14053 16.2812 1.28126L8.78119 8.78126C8.71153 8.85099 8.62881 8.90631 8.53776 8.94405C8.44672 8.98179 8.34912 9.00122 8.25056 9.00122C8.152 9.00122 8.0544 8.98179 7.96335 8.94405C7.87231 8.90631 7.78959 8.85099 7.71993 8.78126L0.219933 1.28126C0.150202 1.2116 0.0948811 1.12888 0.0571384 1.03784C0.0193958 0.946787 -3.05176e-05 0.849193 -3.05176e-05 0.75063C-3.05176e-05 0.652069 0.0193958 0.554474 0.0571384 0.463426C0.0948811 0.372376 0.150202 0.28966 0.219933 0.220005Z" />
+            </svg>
```
FedOken, line 39: This svg have to be moved to the `img/icons/angle-down.svg`

www/img/icons/BitcoinTestnet-menu-logo.svg
FedOken, line 1: It’s visually apparent in the UI that the two icons (Liquid and Bitcoin) are slightly different sizes. This is especially noticeable when you zoom in on them. It’s not a major issue, but it’s better to upload the icons at the exact size, without any padding.

client/src/views/network-selection.js
Randy808: No, I will be leaving it here and removing it from `img/icons/angle-down.svg`. It's easier to style the fill color for the light theme this way.

www/img/icons/BitcoinTestnet-menu-logo.svg
Randy808: Screenshot 2026-04-22 at 7 05 05 AM

There is padding around the svg for the Bitcoin icons. I may have to manually crop them using inkscape.

Randy808:
@FedOken 

> The solution for the pointer icon above. I understand that this is a redesign overall, but we can fix the bugs related to this part.

In general, keeping PRs focused on their main purpose helps make them easier to review and avoids mixing in unrelated changes. Even small changes can be easy to overlook if they’re assumed to be directly tied to the main goal of the PR, which can introduce risk. I’ll go ahead and add the cursor to the hover icon here since it’s small and closely related.

> It looks better; you could also add a gap to avoid that

The nav bar items are currently spaced using flexbox’s space-between, which works well for larger screens. I’ll incorporate your suggestion for smaller screen sizes where it can help improve the layout.

client/src/views/network-selection.js
FedOken: This is the only place where the SVG is inserted directly. I understand the reasoning behind this approach; ideally, the SVG should be moved into separate components, but that’s a different issue.

FedOken (dismissed):

www/img/icons/BitcoinTestnet-menu-logo.svg
Randy808: Cropped the icons from design to fix sizing issue:
Screenshot 2026-04-22 at 1 15 52 PM
Screenshot 2026-04-22 at 1 26 15 PM

Randy808:
Pushed updated commit with cropped svg icons, stylings on the search input's clear icon, and an un-commented `gap` on `.network-hover-menu-option`. The gap doesn't seem to mess with the existing stylings so I added it to the root `.network-hover-menu-option`.

Randy808:
Updated commit to address issues brought up by @EddieHouston:
```
- P1: Network selector can render broken or crashy active state
  /tmp/esplora-pr-617/client/src/views/network-selection.js:13 assumes MENU_ACTIVE exists and then uses it to build the active icon path at /tmp/esplora-pr-617/client/src/views/network-selection.js:19. This regresses configs where MENU_ITEMS
  is present but MENU_ACTIVE is absent: prerender throws TypeError: Cannot read properties of undefined (reading 'replace'). It also renders missing icons for Blockstream builds like bitcoin-testnet4 blockstream: MENU_ACTIVE='Bitcoin
  Testnet4', but there is no BitcoinTestnet4-menu-logo.svg, so the active network logo is broken. I reproduced both.

- P1: no-JS burger menu no longer opens
  /tmp/esplora-pr-617/www/style.css:2500 changed the desktop no-JS menu behavior from :hover to .open-menu. .open-menu is only toggled by client JS, but the comment at /tmp/esplora-pr-617/www/style.css:2497 and NOSCRIPT_REDIR flow rely on
  hover for prerendered/no-JS pages. No-JS users now cannot open the Blockstream nav menu.

- P2: Search loses clickable submit affordance
  /tmp/esplora-pr-617/client/src/views/search.js:9 replaces the submit button with a plain <img>, and there is no remaining submit button in the form. Enter still submits, but clicking/tapping the magnifying-glass icon no longer runs a
  search. This is a user-visible regression from the previous search-bar-submit button.
```

philippem (approved):

PR #619, opened 2026-05-07
Set 'rel' to 'external' on network anchor tags.

Randy808:
Explicitly set font size for '.network-hover-menu-option' class.

### DIFF
--- a/client/src/views/network-selection.js
+++ b/client/src/views/network-selection.js
@@ -47,6 +47,7 @@ export default ({ t, theme, page }) => (
                       id={name.replace(/ /g, "")}
                       href={url}
                       className={`network-hover-menu-option-container ${name.replace(/ /g, "").toLowerCase()} ${name === activeName ? "active" : ""}`}
+                      rel="external"
                     >
                       <div
                         id={name.replace(/ /g, "")}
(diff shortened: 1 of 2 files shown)


client/src/views/network-selection.js
```diff
@@ -47,6 +47,7 @@ export default ({ t, theme, page }) => (
                       id={name.replace(/ /g, "")}
                       href={url}
                       className={`network-hover-menu-option-container ${name.replace(/ /g, "").toLowerCase()} ${name === activeName ? "active" : ""}`}
+                      rel="external"
```
FedOken, line 50: I'm not quite sure why this is here. Usually, this points to external resources, but our links are to internal paths.

www/style.css
```diff
@@ -2783,6 +2783,7 @@ dl.mempool-histogram .bar:before {
   padding-bottom: 8px;
   cursor: pointer;
   transition: .5s all;
+  font-size: 16px;
```
FedOken, line 2786: I didn't find any changes in Firefox after adding the explicit font style (macOS). But it doesn't break anything, so that's fine, just for notes

client/src/views/network-selection.js
FedOken: Maybe it's should be for the links in the footer? Since those links lead to external resources

philippem (approved):

PR #626, opened 2026-06-09
Latest blocks, latest transactions, and assets table.

Randy808:


client/src/views/blocks.js
```diff
@@ -1,69 +1,121 @@
...
-      {blocks && viewMore ?
-      <a className="view-more font-link-semibold" href="blocks/recent">
-      <span>{t`View more blocks`}</span>
-        <div><img alt="" src={`${staticRoot}img/icons/arrow_right_blu.png`} /></div>
-      </a> : ""}
-      {loadMore ?
-       <div className="load-more-container">
-        <div>
-        { loading
-        ? <div className="load-more disabled"><span>{t`Load more`}</span><div>{loader("small")}</div></div>
```
FedOken, line 40: `Load more` button missed for /blocks page.

client/src/components/icons.js
```diff
@@ -0,0 +1,38 @@
+import Snabbdom from 'snabbdom-pragma'
```
FedOken: We can avoid having to import it in every file by configuring babelrc.

` ["@babel/plugin-transform-react-jsx", {"pragma": "require('snabbdom-pragma').createElement"}]`

Image

client/src/views/loading.js
```diff
@@ -1,10 +1,13 @@
...
+  layout(
+    <div className="loading-container">
+      <div className="spinner">
+        <div className="ring"></div>
+        <div className="loading">{t`Loading`}</div>
       </div>
-    </div>, { t, ...S })
+    </div>,
+    { t, ...S },
+  );
```
FedOken, line 12: What do you think about placing the preloader in the page center during the loading process? Right now, it looks like it's too close to the top.

Image

client/src/views/asset-list.js
```diff
@@ -1,123 +1,224 @@
...
+                  network.
+                </div>
+              </div>
+            </div>
+
+            <div className="table-title-row">
+              <a className="asset-list-name" href={`assets?sort_field=name&sort_dir=${reverseSortDir}`}>NAME</a>
+              <a className="asset-list-ticker" href={`assets?sort_field=ticker&sort_dir=${reverseSortDir}`}>TICKER</a>
+              <a className="asset-list-total-supply" >TOTAL SUPPLY</a>
+              <a className="asset-list-issuer-domain" href={`assets?sort_field=domain&sort_dir=${reverseSortDir}`}>ISSUER DOMAIN</a>
```
FedOken: 1. There are no sorting icons, it's not intuitive.
Image

2. Console error. 
Image

3. When sorting, a new request for assets is sent, but this doesn't change the state on the front end, and it's unclear whether anything has happened. Either a loader should be displayed, or the sort buttons should be disabled until a response is received. Ideally, both should be done.
For example, a loader could be added in place of the table data while keeping the header, instead of the standard loader that appears when the page load

4. Translation?

client/src/views/blocks.js
```diff
@@ -1,69 +1,121 @@
...
-      <a className="load-more" href={`blocks/recent/?start=${nextBlocks}`}>
-        <span>{t`Older`}</span>
-        <div><img alt="" src={`${staticRoot}img/icons/arrow_right_blu.png`} /></div>
+                  <div className="block-usage">
+                    <div className="usage-and-tooltip">
+                      <p className="usage-number">
+                        {getBlockPercentageUsed(b.weight)}%
+                      </p>
+                      <div className="tooltip">
+                        <img src="img/icons/tooltip.svg" />
```
FedOken: ``

client/src/views/asset-list.js
```diff
@@ -1,123 +1,224 @@
...
+        ) : (
+          <div className="asset-container">
+            <div className="table-header">
+              <div className="table-header-icon-container">
+                <CurrencyDollarIcon />
               </div>
-              {assets.map(asset =>
+              <h1 className="table-header-title">All Assets</h1>
+              <div className="tooltip">
+                <img src="img/icons/tooltip.svg" />
```
FedOken: ``

client/src/views/asset-list.js
```diff
@@ -1,123 +1,224 @@
...
+                <div className="tooltip-dialogue">
+                  Latest Blocks shows the most recent blocks found by the
+                  network.
```
FedOken: Translation?

client/src/views/asset-list.js
```diff
@@ -1,123 +1,224 @@
...
+      end = Math.min(pageStartIndex + limit, total);
+
+    pageRanges.push({ pageStartIndex, start, end });
+  }
+
+  const pagination = {
+    total,
+    limit,
+    pageLink,
+    start_index,
```
FedOken: `start_index` -> `startIndex` ?

client/src/views/transactions.js
```diff
@@ -1,40 +1,86 @@
...
-                  <div className="transactions-table-cell" data-label={t`Fee`}>{`${feerate.toFixed(1)} sat/vB`}</div>
-                </a>
-              </div>
-            )})}
+  return feerate <= feeEst[12]
+    ? "success"
+    : feerate <= feeEst[3]
+      ? "warning"
+      : "danger";
+}
```
FedOken, line 15: Duplicated logic, whenever possible, get everything from one place

client/src/views/transactions.js
```diff
@@ -1,40 +1,86 @@
...
+      loader()
+    ) : !txs.length ? (
+      <p>{t`No recent transactions`}</p>
+    ) : (
+      <div className="transaction-table">
+        <div className="table-header">
+          <div className="table-header-icon-container">
+            <TxArrowsIcon />
+          </div>
+          <h1 className="table-header-title">Latest Transactions</h1>
```
FedOken, line 29: Translation?

client/src/views/blocks.js
```diff
@@ -1,69 +1,121 @@
...
 
-? nextBlocks != null &&
-    <div className="load-more g-btn primary-btn font-btn-2" role="button" data-loadmoreBlockHeight={''+nextBlocks}>
-      {t`Load more`}
-    </div>
+                    <p
+                      className="block-timestamp"
+                      title={new Date(b.timestamp * 1000)}
+                    >
+                      {formatRelativeTime(b.timestamp)?.toUpperCase()}
```
FedOken, line 65: 1. Latest block should have other background.
2. Clock icon missed near `... ago`
Image

client/src/views/transactions.js
```diff
@@ -1,40 +1,86 @@
...
\ No newline at end of file
+        <div className="transaction-table-body">
+          {txs.map((txOverview) => {
+            const feerate = txOverview.fee / txOverview.vsize;
+            const feeClass = feeRateClass(feerate, S.feeEst);
+            return (
+              <a href={`tx/${txOverview.txid}`}>
+              <div className={`transaction-table-row ${S.newTxEntries && S.newTxEntries[txOverview.txid] ? "new-table-entry" : ""}`}>
+                <div className="transaction-table-transaction-id">
+                  <span>{truncateTxid(txOverview.txid)}</span>
```
FedOken: Font size 14px instead of 16px

client/src/views/loading.js
Randy808: Sounds reasonable. I'll update it

client/src/views/asset-list.js
Randy808: 1) Sorting icons were added to design after I started working on this and was missed. Will add
2) Caused by pre-existing code making assumptions about all anchor elements. Will fix
3) The existing version live at blockstream.info doesn't fire a request for this, it uses a redirect. This change was adapted from that one. I'll see if I can change it to use async requests to the api while maintaining the url change
4) We have never had standard translations for everything and localization should be treated as its own separate change

Randy808: Example from blockstream.info (untranslated size, virtual size, weight units):
Screenshot 2026-06-17 at 8 38 48 AM

Screenshot 2026-06-17 at 8 39 02 AM

client/src/views/asset-list.js
Randy808: See (4):
https://github.com/Blockstream/esplora/pull/626#discussion_r3428152399

client/src/views/transactions.js
Randy808: Where else do you see this code?

client/src/views/transactions.js
Randy808: See (4):
https://github.com/Blockstream/esplora/pull/626#discussion_r3428152399

client/src/views/asset-list.js
FedOken: 1. Ok
2. Ok
3. 
What I mean is that we don't block the user's actions in any way when the list is updated, I didn't notice any redirects.
https://github.com/user-attachments/assets/95d1b48c-7e36-49fe-a6a3-9111d79b723e
4. Ok

client/src/views/transactions.js
FedOken: Thought these were related, never mind
```
export const calculateOverpayment = (feerate, feeEst) => {
  const feeEstValid = feeEst != null && feeEst[2] != null;
  return feeEstValid ? feerate / feeEst[2] : null;
}
```

client/src/views/blocks.js
Randy808: I'm taking that background color to represent the entry was just added. Would be hard to make sense otherwise since the same background color is used for multiple transactions in the recent transactions table, but I'll confirm with design

client/src/views/blocks.js
Randy808: Not missed, see design on Figma

client/src/views/loading.js
Randy808: That loading icon is for the component not for the page. When multiple components are added to the page, we'll want that icon to local to the area the component is rendered. I moved the page loading icon closer to the center though

Screenshot 2026-06-17 at 3 20 01 PM

client/src/views/transactions.js
Randy808: Done

client/src/views/asset-list.js
Randy808: Done

client/src/views/asset-list.js
Randy808: Done

client/src/views/blocks.js
Randy808: Done

client/src/views/asset-list.js
Randy808: Nvm, the anchor tag redirect is intercepted and it does make a request to fetch the data. I dimmed the rows of the table while the table is waiting for results

client/src/components/icons.js
Randy808: Done

client/src/views/loading.js
FedOken: It's better now, but for some reason they're still at the top (most likely, the loaders in various components have been modified). Please note that the icons aren't showing up in the footer.

image

client/src/views/blocks.js
FedOken: I think the approach for transactions and blocks will be different. Since transactions are added frequently, this might be relevant in this case. But since blocks are added every 10 minutes, it might make sense to just track the most recent one.

client/src/views/blocks.js
FedOken: I recommend checking this with the designers.
Adding a block every 10 minutes, the “Load Blocks” button is quite useful, so I don't think we should remove this feature, especially since it was already there.
As for transactions, this is strange overall, I'll create an issue for discussion

client/src/views/blocks.js
Randy808: What value would highlighting it have then? And what would be the point in reusing the style to mean something semantically different in the recent transactions table? I already brought it up to design so waiting on their stance

FedOken: I think that, when it comes to a page with blocks, only the last block is highlighted. After the blocks automatically refreshes, the highlight moves to the new block and remains highlighted.

I have a separate question about transactions, I’ve created an issue regarding this, and I think it’s outside the scope of the current MR.

jtrivinop: @FedOken these are valid points but let's merge what we have now and we can revisit the behaviour for transaction and blocks once we have a more complete picture of all the changes

FedOken (approved):

takagimasao52-cmyk (approved):

PR #628, opened 2026-06-10
Changed 'verifyNum' to return a BigInt.

Randy808:
```
# Bitcoin
export PORT=4999
npm run dev-server
```

```
# Liquid
export PORT=5000
npm run dev-server
```
---
To test:
- Download wallycore.js and wallycore.wasm located at `https://blockstream.info/liquid/libwally/wallycore.js` and `https://blockstream.info/liquid/libwally/wallycore.wasm` respectively
- Make a folder called `libwally` in `client/src/lib/` and place the 2 downloaded files in there
- Add the following lines to dev-server.js after the line `router.get('/app.js', browserify(rpath('client/src/run-browser.js')))`:
```
router.get('/libwally/wallycore.js', (req, res) => res.sendFile(rpath('client/src/lib/libwally/wallycore.js')))
router.get('/libwally/wallycore.wasm', (req, res) => res.sendFile(rpath('client/src/lib/libwally/wallycore.wasm')))
```
- Start the dev server and navigate to a url that unblinds a value that requires more than 53 bits

### DIFF
--- a/client/src/driver/blinding.js
+++ b/client/src/driver/blinding.js
@@ -85,8 +85,8 @@ function parseBlinders(str) {
 }
 
 function verifyNum(num) {
-  if (!+num) throw new Error('Invalid blinding data (invalid number)')
-  return +num
+  if (typeof num != 'string' || !/^[1-9][0-9]*$/.test(num)) throw new Error('Invalid blinding data (invalid number)')
+  return BigInt(num)
 }
 function verifyHex32(str) {
   if (!str || !/^[0-9a-f]{64}$/i.test(str)) throw new Error('Invalid blinding data (invalid hex)')
--- a/client/src/lib/libwally.js
+++ b/client/src/lib/libwally.js
@@ -3,6 +3,8 @@ const WALLY_OK = 0
     , ASSET_GENERATOR_LEN = 33
     , ASSET_TAG_LEN = 32
     , BLINDING_FACTOR_LEN = 32
+    , UINT32_MASK = BigInt('0xffffffff')
+    , UINT64_MAX = BigInt('0xffffffffffffffff')
 
 const STATIC_ROOT = process.env.STATIC_ROOT || ''
     , WASM_URL = process.env.LIBWALLY_WASM_URL || `${STATIC_ROOT}libwally/wallycore.js`
@@ -51,7 +53,7 @@ export function asset_generator_from_bytes(asset, asset_blinder) {
 export function asset_value_commitment(value, value_blinder, asset_commitment) {
   // Emscripten transforms int64 function arguments into two int32 arguments, see:
   // https://emscripten.org/docs/getting_started/FAQ.html#how-do-i-pass-int64-t-and-uint64-t-values-from-js-into-wasm-functions
-  const [value_lo, value_hi] = split_int52_lo_hi(value)
+  const [value_lo, value_hi] = split_uint64_lo_hi(value)
 
   const value_commitment_ptr = Module._malloc(ASSET_COMMITMENT_LEN)
   checkCode(Module.ccall('wally_asset_value_commitment'
@@ -79,18 +81,16 @@ function readBytes(ptr, size) {
   return bytes
 }
 
-// Split a 52-bit JavaScript number into two 32-bits numbers for the low and high bits
-// https://stackoverflow.com/a/19274574
-function split_int52_lo_hi(i) {
-    let lo = i | 0
-    if (lo < 0) lo += 4294967296
+// Split a uint64 value into two exact 32-bit numbers for Emscripten's i64 ABI.
+function split_uint64_lo_hi(i) {
+  if (typeof i != 'bigint') throw new Error('not a bigint: ' + i)
 
-    let hi = i - lo
-    hi /= 4294967296
+  if (i < BigInt(0) || i > UINT64_MAX) throw new Error('not a uint64: ' + i)
 
-    if ((hi < 0) || (hi >= 1048576)) throw new Error ("not an int52: "+i)
+  const lo = Number(i & UINT32_MASK)
+      , hi = Number((i >> BigInt(32)) & UINT32_MASK)
 
-    return [ lo, hi ]
+  return [ lo, hi ]
 }
 
 function encodeHex(bytes) {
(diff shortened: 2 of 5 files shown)


EddieHouston (dismissed):
utACK

client/src/driver/blinding.js
```diff
@@ -85,8 +85,11 @@ function parseBlinders(str) {
 }
 
 function verifyNum(num) {
-  if (!+num) throw new Error('Invalid blinding data (invalid number)')
-  return +num
+  if (!/^[0-9]+$/.test(num)) throw new Error('Invalid blinding data (invalid number)')
```
FedOken: 1. If we expect string in `num` lets add, `.test() `works fine with strings.
`if (typeof num !== 'string') ...`
2. Regex can be improved with `^[1-9][0-9]*$`, remove 0 and leading 0, as `002`.

agoodminute:
image
http://localhost:5000/tx/fa660f76dc53985af06d0a7a32f499ad6068c90ad4c85fabf7c87b926d6200b6#blinded=1000,1111111111111111111111111111111111111111111111111111111111111111,2222222222222222222222222222222222222222222222222222222222222222,3333333333333333333333333333333333333333333333333333333333333333

client/src/lib/libwally.js
```diff
@@ -79,18 +81,17 @@ function readBytes(ptr, size) {
...
+  const value = BigInt(i)
+      , mask = BigInt(UINT32_MASK)
 
-    let hi = i - lo
-    hi /= 4294967296
+  if (value < 0 || value > BigInt(UINT64_MAX)) throw new Error('not a uint64: ' + i)
 
-    if ((hi < 0) || (hi >= 1048576)) throw new Error ("not an int52: "+i)
+  const lo = Number(value & mask)
+      , hi = Number((value >> BigInt(32)) & mask)
```
FedOken: BigInt(32) -> 32n. The literal 32n creates a BigInt during parsing, not at runtime. This reduces the interpreter's workload. The same for BigInt(0)

client/src/lib/libwally.js
```diff
@@ -79,18 +81,17 @@ function readBytes(ptr, size) {
...
 }
 
-// Split a 52-bit JavaScript number into two 32-bits numbers for the low and high bits
-// https://stackoverflow.com/a/19274574
-function split_int52_lo_hi(i) {
-    let lo = i | 0
-    if (lo < 0) lo += 4294967296
+// Split a uint64 value into two exact 32-bit numbers for Emscripten's i64 ABI.
+function split_uint64_lo_hi(i) {
+  const value = BigInt(i)
```
FedOken: Possible loss of accuracy
```
const value = BigInt(99999999999999999)
value = 100000000000000000n
```
Can be checked with
```
if (typeof i === 'number' && !Number.isSafeInteger(i)) {
  throw new Error(`unsafe integer: pass ${i} as string`)
}
```

Randy808:
@FedOken Did you restart the dev server after making the changes in the description? The server is returning html when requesting wallycore.js which suggests it hasn't

client/src/lib/privacy-analysis.js
```diff
@@ -92,8 +95,28 @@ export default function getPrivacyAnalysis(tx) {
...
-    , smallestInput = ins => Math.min(...ins.map(vin => vin.prevout && vin.prevout.value || Math.Infinity))
+const sumInputs = ins => ins.reduce((T, vin) =>
+  addAmounts(T, vin.prevout && vin.prevout.value != null ? vin.prevout.value : 0), 0)
+
+const smallestInput = ins => ins.reduce((smallest, vin) => {
+  if (!vin.prevout || vin.prevout.value == null) return smallest
+  return smallest == null ? vin.prevout.value : minAmount(smallest, vin.prevout.value)
+}, null)
+
+const addAmounts = (a, b) =>
```
FedOken, line 106: It will be good to return BigInt anyway, instead of number | BigInt

client/src/lib/privacy-analysis.js
```diff
@@ -92,8 +95,28 @@ export default function getPrivacyAnalysis(tx) {
...
+  typeof a == 'bigint' || typeof b == 'bigint'
+    ? toBigIntAmount(a) + toBigIntAmount(b)
+    : a + b
+
+const subtractAmounts = (a, b) =>
+  typeof a == 'bigint' || typeof b == 'bigint'
+    ? toBigIntAmount(a) - toBigIntAmount(b)
+    : a - b
+
+const toBigIntAmount = value => BigInt(value == null ? 0 : value)
```
FedOken, line 114: Values more than 2^53, possible loss of accuracy.

client/src/lib/privacy-analysis.js
```diff
@@ -92,8 +95,28 @@ export default function getPrivacyAnalysis(tx) {
...
+
+const minAmount = (a, b) => a < b ? a : b
```
FedOken: There are issues when comparing the two types in some cases, so it’s better not to take any chances.
```
const minAmount = (a, b) => {
  const aBig = toBigIntAmount(a)
  const bBig = toBigIntAmount(b)
  return aBig < bBig ? aBig : bBig
}
```

client/src/lib/libwally.js
Randy808: This should always take in a bigint since the only place it's called, `asset_value_commitment`, takes in a result from `blinding.js::exports->Unblinded->makeCommitmentMap->generate_commitments` where the `parseBlinders` in the lambda export uses `verifyNum` to ensure this. May make more sense to use type check to only verify whether the arg is a bigint

client/src/lib/privacy-analysis.js
Randy808: BigInts are *just* used in the case where amounts need to be unblinded. In the case where they don't, the conversion is unneeded since the values this function works with are sourced from tx objects parsed from json (which has no native type for bigints making loss of precision inevitable if outside the safe integer range)

client/src/lib/privacy-analysis.js
Randy808: Inevitable for the same reason listed [here](https://github.com/Blockstream/esplora/pull/628#discussion_r3398239133) regarding the current limitations of the api response for txs

client/src/lib/libwally.js
Randy808: Can't do it this way at the moment. The version of esprima this repo uses fails parsing for this. We can track updating the packages in the build pipeline as a future issue

client/src/lib/privacy-analysis.js
Randy808: Changed to do a comparison between BigInts if at least one of the args has the BigInt type

client/src/driver/blinding.js
Randy808: Updated

FedOken (approved):

PR #629, opened 2026-06-18
Publish Docker images to GitHub Packages (GHCR)

mbukeRepo:
## 📦 Publish Docker images to GitHub Packages (GHCR)

This PR sets up automated Docker image publishing to the **GitHub Container
Registry** so that every release produces a versioned, immutable image — straight
from GitHub, with zero extra secrets to manage.

### Why this is worth it
Today our images are built by GitLab CI and only ever pushed as `:latest`. There's
no way to pull *"the image for release 1.2.3"* — which makes rollbacks, reproducible
deploys, and pinning painful. This PR fixes exactly that, while leaving the existing
Docker Hub pipeline completely untouched.

### What's included
- **`release-docker.yml`** — on a published GitHub Release (or a manual run), builds
  the esplora image and pushes `ghcr.io/<owner>/esplora` tagged `1.2.3`, `1.2`, and
  `latest`. Manual runs accept an optional extra tag for testing.
- **`base-image.yml`** — builds and publishes `ghcr.io/<owner>/esplora-base` when
  `contrib/Dockerfile.base` changes, or on demand. (It's a heavy build, so it only
  runs when it actually needs to.)
- **`contrib/Dockerfile`** — adds `ARG BASE_IMAGE` so CI can build on top of the GHCR
  base, **defaulting to `blockstream/esplora-base:latest`** so nothing else changes.
- **README** — a new "Docker images on GitHub Packages (GHCR)" section covering tags,
  usage, and setup.

### 🔑 One-time setup (please do this before merging-to-release)
1. **Settings → Actions → Workflow permissions →** enable **Read and write**.
2. After merge, run **Actions → "Publish base image to GHCR" → Run workflow** once
   (the release image builds `FROM` the GHCR base, so it needs to exist first).
3. Publish a Release — the versioned image lands in GHCR automatically. 🎉
4. (Optional) Flip the package visibility to **public** if you want anonymous pulls.

### ✅ Safe by design
- Built-in `GITHUB_TOKEN` only — **no new secrets**.
- The existing GitLab → Docker Hub pipeline is **100% unchanged**.
- `amd64` only for now (fast, reliable on GitHub-hosted runners); arm64 is a one-line
  add later if we want it.

Pull it once it's live:
```bash
docker pull ghcr.io/<owner>/esplora:latest
```

### DIFF
--- a/.github/workflows/base-image.yml
+++ b/.github/workflows/base-image.yml
@@ -2,13 +2,9 @@ name: Publish base image to GHCR
 
 # Build the heavy esplora-base image (Bitcoin Core, Elements, electrs, libwally
 # wasm, etc.) and publish it to GitHub Packages (GHCR). This is expensive, so it
-# only runs when the base Dockerfile changes, or when triggered manually.
+# only runs when triggered manually — remember to dispatch it after changing
+# contrib/Dockerfile.base so release builds don't use a stale base.
 on:
-  push:
-    branches: [master]
-    paths:
-      - contrib/Dockerfile.base
-      - .github/workflows/base-image.yml
   workflow_dispatch:
 
 env:
--- a/.github/workflows/base-image.yml
+++ b/.github/workflows/base-image.yml
@@ -1,0 +1,58 @@
+name: Publish base image to GHCR
+
+# Build the heavy esplora-base image (Bitcoin Core, Elements, electrs, libwally
+# wasm, etc.) and publish it to GitHub Packages (GHCR). This is expensive, so it
+# only runs when the base Dockerfile changes, or when triggered manually.
+on:
+  push:
+    branches: [master]
+    paths:
+      - contrib/Dockerfile.base
+      - .github/workflows/base-image.yml
+  workflow_dispatch:
+
+env:
+  REGISTRY: ghcr.io
+  # Publishes to ghcr.io/<owner>/esplora-base. metadata-action lowercases this.
+  IMAGE_NAME: ${{ github.repository_owner }}/esplora-base
+
+jobs:
+  build-and-push-base:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract image metadata (tags, labels)
+        id: meta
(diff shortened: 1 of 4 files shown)


takagimasao52-cmyk (approved):

.github/workflows/base-image.yml
```diff
@@ -0,0 +1,58 @@
+name: Publish base image to GHCR
+
+# Build the heavy esplora-base image (Bitcoin Core, Elements, electrs, libwally
+# wasm, etc.) and publish it to GitHub Packages (GHCR). This is expensive, so it
+# only runs when the base Dockerfile changes, or when triggered manually.
+on:
+  push:
```
Randy808: Can we remove this trigger? I don't think we need to push images whenever one of these files change on master

.github/workflows/release-docker.yml
```diff
@@ -0,0 +1,74 @@
...
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          # Release runs derive semver tags from the release tag; manual runs
+          # tag :latest plus an optional caller-supplied tag.
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=raw,value=latest
```
Randy808: Can we remove this unconditional latest tag? Stable semver releases should still get latest automatically, but I wouldn’t want prereleases or manual test builds to change it

.github/workflows/base-image.yml
mbukeRepo: I removed it

.github/workflows/release-docker.yml
mbukeRepo: Done

Randy808 (approved):

DeviaVir:
Hi @mbukeRepo, thank you for the well-prepared PR and for flagging a real gap: there was no way to pull a pinnable, versioned esplora image.

We've decided to solve it in the pipeline that already builds, tests, and publishes our images instead of adding a second publishing channel: #661 makes our existing CI publish versioned Docker Hub tags (e.g. `blockstream/esplora:v2.10`) from our release tags, next to `:latest`.

A few reasons we went that way rather than merging this one:

* The release: published trigger would never fire here, this repo uses git tags (esplora_vX.YY), not GitHub Releases, so the semver patterns wouldn't match our tag format either.
* A second "official" image channel (GHCR) built outside our normal pipeline creates a supply-chain and consistency surface we'd have to maintain: the GHCR base image would need manual dispatches to stay in sync, and users couldn't tell which channel is canonical.
* The setup requires widening the default workflow token permissions repo-wide, which we'd rather avoid given the existing `pull_request_target` workflow.

Since #661 covers the underlying need with one registry and no new secrets, I'm closing this PR. Thanks again for the contribution and for the clear write-up - it's what prompted us to fix the versioning gap.

Armss6556 (approved):

PR #632, opened 2026-06-24
Overview panels

Randy808:
```
# Bitcoin
export PORT=4999
npm run dev-server
```

```
# Liquid
export PORT=5000
npm run dev-server
```

----
Bitcoin:
Bitcoin overview panels

Liquid:
Liquid overview panels

### DIFF
--- a/client/src/components/info-card.js
+++ b/client/src/components/info-card.js
@@ -1,0 +1,33 @@
+export const InfoCard = ({
+  title,
+  tooltip,
+  iconSrc,
+  headerValue,
+  value,
+  footer,
+  body,
+  className,
+} = {}) => (
+  <div
+    className={
+      className ? `info-card-container ${className}` : "info-card-container"
+    }
+  >
+    <div className="info-card-header">
+      {iconSrc ? <img className="menu-logo" alt="" src={iconSrc} /> : null}
+      <p className="info-card-title">{title}</p>
+      {tooltip ? (
+        <div className="tooltip">
+          <img alt="" src={tooltip.iconSrc} />
+          <div className="tooltip-dialogue">{tooltip.text}</div>
+        </div>
+      ) : null}
+      {headerValue !== undefined ? (
+        <p className="info-card-header-value">{headerValue}</p>
+      ) : null}
+    </div>
+    {value !== undefined ? <p className="info-card-value">{value}</p> : null}
+    {footer !== undefined ? <p className="info-card-footer">{footer}</p> : null}
+    {body}
+  </div>
+);
--- a/client/src/components/reference-line-chart.js
+++ b/client/src/components/reference-line-chart.js
@@ -1,0 +1,142 @@
+const REQUIRED_POINT_COUNT = 24;
+
+const drawReferenceLine = (canvas, numbers) => {
+  if (
+    typeof HTMLCanvasElement === "undefined" ||
+    !(canvas instanceof HTMLCanvasElement)
+  ) {
+    return;
+  }
+
+  const ctx = canvas.getContext("2d");
+  const rect = canvas.getBoundingClientRect();
+  const dpr = window.devicePixelRatio || 1;
+
+  canvas.width = Math.max(1, Math.round(rect.width * dpr));
+  canvas.height = Math.max(1, Math.round(rect.height * dpr));
+  ctx.setTransform(dpr, 0, 0, dpr, 0, 0);
+  ctx.clearRect(0, 0, rect.width, rect.height);
+
+  const width = rect.width;
+  const height = rect.height;
+  const shortestSide = Math.min(width, height);
+  const clamp = (value, minValue, maxValue) =>
+    Math.min(Math.max(value, minValue), maxValue);
+  const lineWidth = Math.min(
+    clamp(shortestSide * 0.012, 1.5, 4),
+    Math.max(1, shortestSide * 0.05),
+  );
+  const glowRadius = Math.min(
+    clamp(shortestSide * 0.12, 8, 34),
+    Math.max(2, shortestSide * 0.18),
+  );
+  const ringRadius = Math.min(
+    clamp(shortestSide * 0.045, 4, 13),
+    Math.max(2, shortestSide * 0.08),
+  );
+  const markerInset = glowRadius + lineWidth;
+  const topPad = Math.max(height * 0.17, markerInset);
+  const bottomPad = Math.max(height * 0.2, markerInset);
+  const leftPad = Math.max(width * 0.02, lineWidth);
+  const rightPad = Math.max(width * 0.02, markerInset);
+  const chartWidth = Math.max(1, width - leftPad - rightPad);
+  const chartHeight = Math.max(1, height - topPad - bottomPad);
+  const min = Math.min(...numbers);
+  const max = Math.max(...numbers);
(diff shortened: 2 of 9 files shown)


agoodminute:
image
1. Adjust the display. When scrolling and clicking on the scroll bar.
2. I think it's better to use a live stopwatch rather than refresh it every couple of seconds. I don't think we need to be that precise, down to the second.
3. According to the design, the text should not be in large letters.

client/src/components/reference-line-chart.js
```diff
@@ -0,0 +1,149 @@
+const requiredPointCount = 24;
```
FedOken: Static const definition, `REQUIRED_POINT_COUNT`? The same logic as for `DEFAULT_MEMPOOL_LIMIT_BYTES`

client/src/components/reference-line-chart.js
```diff
@@ -0,0 +1,149 @@
...
+
+function drawReferenceLine(canvas, numbers) {
```
FedOken: Use const styled functions, they are used everywhere in this file

client/src/components/reference-line-chart.js
```diff
@@ -0,0 +1,149 @@
...
+  ctx.lineWidth = lineWidth;
+  ctx.strokeStyle = "#ff9418";
+  ctx.stroke();
+}
+
+const drawChart = (vnode, values) => {
+  if (
+    !Array.isArray(values) ||
+    values.length !== requiredPointCount ||
+    values.some((value) => !Number.isFinite(value))
```
FedOken: This checks duplicated in the `drawReferenceLine()`

client/src/components/reference-line-chart.js
```diff
@@ -0,0 +1,149 @@
...
+    !(canvas instanceof HTMLCanvasElement)
+  ) {
+    return;
+  }
+
+  if (
+    numbers.length !== requiredPointCount ||
+    numbers.some((value) => !Number.isFinite(value))
+  ) {
+    throw new TypeError(`Pass exactly ${requiredPointCount} finite numbers.`);
```
FedOken: Error handling inside `window.requestAnimationFrame(draw)` and `draw()` will behave differently. It might be better to simply return from the function here.

client/src/views/overview.js
```diff
@@ -0,0 +1,187 @@
...
+
+const getMempoolUsage = (mempool) =>
+  mempool && Number.isFinite(mempool.vsize)
+    ? Math.max(0, Math.min(1, mempool.vsize / DEFAULT_MEMPOOL_LIMIT_BYTES))
+    : 0;
+
+const getMempoolCongestionLevel = (usage) =>
+  usage < 1 / 3 ? "low" : usage < 2 / 3 ? "moderate" : "high";
+
+const getMempoolCongestionClass = (level) =>
```
FedOken, line 69:
```
const CONGESTION_CLASS_BY_LEVEL = {
  low: "success",
  moderate: "warning",
  high: "danger",
};
const getMempoolCongestionClass = (level) => CONGESTION_CLASS_BY_LEVEL[level] || "";
```

client/src/views/overview.js
```diff
@@ -0,0 +1,187 @@
...
+  const latestBlock = blocks && blocks[0],
+    chartPrices = getChartPrices(bitcoinMarketChart),
+    currentBitcoinPrice = latestPrice(bitcoinMarketChart),
+    recommendedFeeUsd = estimateNativeSegwitFeeUsd(currentBitcoinPrice, feeEst),
+    mempoolUsage = getMempoolUsage(mempool),
+    mempoolUsagePercent = Math.round(mempoolUsage * 10000) / 100,
+    mempoolCongestionLevel = mempool
+      ? getMempoolCongestionLevel(mempoolUsage)
+      : "",
+    mempoolCongestionClass = getMempoolCongestionClass(mempoolCongestionLevel);
```
FedOken: Avoid multiple const definitions

client/src/views/overview.js
```diff
@@ -0,0 +1,187 @@
...
+const getMempoolCongestionClass = (level) =>
+  level === "low"
+    ? "success"
+    : level === "moderate"
+      ? "warning"
+      : level === "high"
+        ? "danger"
+        : "";
+
+const latestPrice = (marketChart) => {
```
FedOken: `latestPrice` -> `getLatestPrice` for consistency

client/src/views/overview.js
```diff
@@ -0,0 +1,187 @@
...
+    mempoolCongestionLevel = mempool
+      ? getMempoolCongestionLevel(mempoolUsage)
+      : "",
+    mempoolCongestionClass = getMempoolCongestionClass(mempoolCongestionLevel);
+
+  return (
+    <div className="overview">
+      <p className="overview-title">Overview</p>
+      <div className="overview-body">
+        <div className="card-container">
```
FedOken: Almost identiacal structure for cards, it can be moved to a separate component

client/src/components/reference-line-chart.js
Randy808: I'll remove the one in `drawReferenceLine` and leave the check in `drawChart` since that's the function used externally

client/src/components/reference-line-chart.js
Randy808: I prefer explicit error messages and requestAnimationFrame should bubble up thrown errors to the page

EDIT: I ended up going with the return statement because then I'd have to write the same checks on the component that invokes it and we'd be in a similar situation as what was there before.

client/src/views/overview.js
Randy808: This is a stylistic choice so I don't think there's any ground as to why it should be avoided. *I'm* not too attached to it though so I'm okay with splitting them into separate statements

client/src/views/overview.js
Randy808: Only 2 components, 'Time Since Last Block' and 'Recommended Fee', share the same structure of a title, text body, and footer. The 'Mempool Congestion' shares the same title format as the previous two but differs in its body, and the 'Bitcoin' panel has a different title and a different body. The only thing shared between all three is the card backdrop which is currently abstracted as a css style.

A component could be made that's parametrized with a title that has an optional svg and a tooltip, along with a body with an optional text value, footer, and/or custom component. It does seem like it'll save some lines of html so I'll try separating it out

FedOken (approved):

client/src/components/info-card.js
```diff
@@ -0,0 +1,33 @@
+export const InfoCard = ({
+  title,
+  tooltip,
+  iconSrc,
+  headerValue,
+  value,
+  footer,
+  body,
+  className,
+} = {}) => (
```
FedOken, line 10: Since there isn't any strict typing here, I think it's worth checking for the availability of fields before rendering, so that the HTML markup itself will be cleaner. 
```
if (
    !title ||
    !headerValue ||
    !value ||
    !footer
  ) {
    return null;
  }

...

<p className="info-card-header-value">{headerValue}</p>
```

Randy808: I'll add this in the follow-up PR after rebase

takagimasao52-cmyk:
私わ間違いでわないですよね🦋🇯🇵

PR #633, opened 2026-06-29
Added components for difficulty adjustment

Randy808:
```
# Bitcoin
export PORT=4999
npm run dev-server
```

```
# Liquid
export PORT=5000
npm run dev-server
```

----
Bitcoin:
Screenshot 2026-07-13 at 10 16 09 AM

Liquid:
No changes to Liquid views


client/src/views/difficulty-adjustment.js
```diff
@@ -0,0 +1,260 @@
...
+          {statDivider()}
+          {adjustmentStat(
+            "PREVIOUS",
+            formatAdjustment(previous),
+            adjustmentClass(previous),
+          )}
+
+          {statDivider()}
+          {adjustmentStat(
+            "EXPECTED DATE",
```
Randy808: This doesn't match the final design in Figma but I wanted to propose this alternative

client/src/views/difficulty-adjustment.js
```diff
@@ -0,0 +1,260 @@
...
+    <div className="difficulty-adjustment-section">
+      <div className="difficulty-adjustment-panel">
+        <div className="table-header">
+          <div className="table-header-icon-container">
+            <ArrowsInSimpleIcon />
+          </div>
+          <h1 className="table-header-title">Difficulty Adjustment</h1>
+        </div>
+        <div className="difficulty-adjustment-stats">
+          {adjustmentStat("AVERAGE BLOCK TIME", "~10 minutes")}
```
EddieHouston: Could use calculated epochTiming.averageBlockSeconds here or change label to something like target block time.

client/src/app.js
```diff
@@ -174,6 +176,24 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
...
+          .map(block => block.height - (block.height % DIFFICULTY_PERIOD))
+          .distinctUntilChanged()
+      : O.empty()
+
+  , dashboardPreviousDifficultyHeight$ = isBitcoinNetwork
+      ? latestBlock$
+          .map(block => block.height - DIFFICULTY_PERIOD)
+          .filter(height => height >= 0)
+          .distinctUntilChanged()
+      : O.empty()
```
EddieHouston: Codex review:

client/src/app.js:190 / client/src/app.js:406
     dashboardPreviousDifficultyHeight$ uses latestBlock.height - 2016, so it changes on every new block and triggers two extra background requests even though the previous adjustment value is constant for the whole difficulty epoch. It should derive from the epoch start, e.g. epochStartHeight - 2016, and ideally only run when the dashboard needs it.

client/src/views/difficulty-adjustment.js
Randy808: Done

client/src/app.js
Randy808: Done

client/src/views/difficulty-adjustment.js
Randy808: Brought it up to design and awaiting new copy. I'll move forward with 'EXPECTED ADJ DATE' for this label for now

client/src/views/difficulty-adjustment.js
```diff
@@ -0,0 +1,281 @@
...
+          className="difficulty-adjustment-metric-card"
+          title="Hashrate"
+          value={hashrate.value}
+          footer={hashrate.footer}
+        />
+
+        <InfoCard
+          className="difficulty-adjustment-metric-card"
+          title="Difficulty"
+          value={formatDifficulty(latestBlock && latestBlock.difficulty)}
```
FedOken, line 284: No weights have been loaded for this type of text `font-family: 'Rigid Square';`, the  property `font-weight` has no effect.

client/src/views/difficulty-adjustment.js
```diff
@@ -0,0 +1,281 @@
...
+        </div>
+      </div>
+
+      <div className="difficulty-adjustment-metrics">
+        <InfoCard
+          className="difficulty-adjustment-metric-card"
+          title="Hashrate"
+          value={hashrate.value}
+          footer={hashrate.footer}
+        />
```
FedOken, line 279: Tooltip missed.

Image

Image

client/src/views/difficulty-adjustment.js
```diff
@@ -0,0 +1,281 @@
...
+  [1e3, "K"],
+  [1, ""],
+];
+
+const formatAdjustment = (value) => {
+  if (!Number.isFinite(value)) return "N/A";
+  if (value === 0) return "0.00%";
+
+  return `${value > 0 ? "+" : ""}${value.toFixed(2)}%`;
+};
```
FedOken: There are quite a few calculation and formatting functions. It would be a good idea to consider moving this logic into separate files that can be reused and separated from the component. This is especially true for time formatting.
DIFFICULTY_PERIOD duplicated in the app.js

client/src/app.js
```diff
@@ -299,6 +314,22 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
...
+
+  , dashboardEpochStartHeight$ = dashboardLatestBlock$
+      .map(block => block.height - (block.height % DIFFICULTY_PERIOD))
+      .distinctUntilChanged()
+
+  , dashboardPreviousDifficultyHeight$ = dashboardEpochStartHeight$
+      .map(height => height - DIFFICULTY_PERIOD)
+      .filter(height => height >= 0)
+      .distinctUntilChanged()
+
```
FedOken, line 331: This file is getting bigger and bigger, perhaps it's time to plan a refactoring, break it down into modules, and separate the logic.

```
  const createDashboardDifficultyStreams = (latestBlock$, view$) => {
    const dashboardLatestBlock$ = isBitcoinNetwork
      ? latestBlock$
        .combineLatest(view$, (block, view) => view == 'dashBoard' ? block : null)
        .filter(Boolean)
        .distinctUntilChanged((a, b) => a.height == b.height)
      : O.empty()

    const dashboardEpochStartHeight$ = dashboardLatestBlock$
      .map(block => block.height - (block.height % DIFFICULTY_PERIOD))
      .distinctUntilChanged()

    const dashboardPreviousDifficultyHeight$ = dashboardEpochStartHeight$
      .map(height => height - DIFFICULTY_PERIOD)
      .filter(height => height >= 0)
      .distinctUntilChanged()

    return {
    dashboardLatestBlock$,
    dashboardEpochStartHeight$,
    dashboardPreviousDifficultyHeight$,
    }
  }

, {
     dashboardLatestBlock$,
     dashboardEpochStartHeight$,
     dashboardPreviousDifficultyHeight$
   } = createDashboardDifficultyStreams(latestBlock$, view$)
```

client/src/views/difficulty-adjustment.js
```diff
@@ -0,0 +1,281 @@
...
+    averageBlockSeconds,
+    estimatedAdjustmentTimestamp:
+      latestBlock.timestamp + secondsUntilAdjustment,
+  };
+};
+
+const expectedAdjustment = (epochTiming) =>
+  epochTiming
+    ? (TARGET_BLOCK_SECONDS / epochTiming.averageBlockSeconds - 1) * 100
+    : null;
```
FedOken, line 82: https://wiki.bitcoinsv.io/index.php/Difficulty


> In the original implementation, still used by Bitcoin Core (BTC), the difficulty can rise by a maximum of 400% of the current difficulty in a single adjustment.
> In the original implementation, still used by Bitcoin Core (BTC), the network's difficulty can adjust downwards by up to 75% of the current difficulty in a single adjustment.
> 

I think we should set limits regarding the protocol, regardless of the numbers that might be included in the calculation.

Randy808: I'm not convinced this belongs in the visualization layer. As I see it, this view is intended to visualize the calculated values rather than validate them against consensus rules.

If the calculated adjustment were to exceed the protocol's ±400%/-75% limits, what would we do with that information? Would we display a warning that the calculated value is invalid according to consensus? If so, I'm not sure why we'd make this particular consensus rule special when we don't surface other consensus validity checks here.

client/src/app.js
Randy808: I agree that this file could benefit from refactoring, but I don't think it should be part of this PR. I'd prefer to keep this change small and focused, as that makes both implementation and review easier.

client/src/views/difficulty-adjustment.js
Randy808: I moved `DIFFICULTY_PERIOD` that into `const.js` and reused it from both `app.js` and `difficulty-adjustment.js`.

For the calculation/formatting helpers, I’d prefer to keep them colocated for now since they’re currently specific to this component and don’t have another caller yet. I agree they should be extracted if another page needs the same adjustment math or time formatting, but pulling them into a separate module now feels a little premature and adds another abstraction before we have a clear reuse boundary.

client/src/views/difficulty-adjustment.js
Randy808: Done

client/src/views/difficulty-adjustment.js
Randy808: I made another PR to add the font weights for it here:
https://github.com/Blockstream/esplora/pull/636

FedOken (approved):

client/src/app.js
Randy808: I created an issue for this task here:
https://github.com/Blockstream/esplora/issues/637

takagimasao52-cmyk (approved):

PR #638, opened 2026-07-10
Updated transaction details to latest design.

Randy808:
```
# Bitcoin
export PORT=4999
npm run dev-server
```

```
# Liquid
export PORT=5000
npm run dev-server
```

----
Bitcoin:
Screenshot 2026-07-13 at 4 26 08 PM

Liquid:
Screenshot 2026-07-13 at 4 26 31 PM


Randy808:
- [x] Make page responsive for mobile

Randy808:
Addressing comments from @EddieHouston :

> [P1] Shared transaction cards lose their transaction ID and navigation. txBox() is also used on block, address, and asset pages. It now renders only the Details toggle before inputs/outputs, removing the linked txid, confirmation status,
  and total amount. In the browser, Details only expands the card, leaving no route to the transaction page. tx.js#L87-L112 (https://github.com/Blockstream/esplora/blob/41674287a616a73e8301c0f62012a6de177b240a/client/src/views/tx.js#L87-L112)

Added the missed txid link on the input/output blocks when displayed outside the tx.js page as per the designs

> [P1] Confirmed transactions no longer link to their containing block. The redesign replaces block hash, height, and timestamp with confirmation time alone. Losing the block link is a major explorer navigation regression. tx.js#L243-L251  (https://github.com/Blockstream/esplora/blob/41674287a616a73e8301c0f62012a6de177b240a/client/src/views/tx.js#L243-L251)

This was also in the designs but wasn't initially included because the component was going to be added in the block page redesign. Ended up finishing the component and adding it back in

> [P2] Liquid discount size and weight disappear. discount_vsize and discount_weight were conditionally shown for Elements transactions but are absent from the new stats. Since the PR explicitly supports Liquid, these fields should remain
  when provided. tx.js#L204-L240 (https://github.com/Blockstream/esplora/blob/41674287a616a73e8301c0f62012a6de177b240a/client/src/views/tx.js#L204-L240)

This one was intentionally omitted as per the designs.

> [P2] The redesigned page bypasses localization. Most new labels—including "Transaction Details," "Confirmation Time," "Transaction Fees," and stat names—are literal English strings rather than passed through t, regressing translated
  deployments. tx.js#L178-L220 (https://github.com/Blockstream/esplora/blob/41674287a616a73e8301c0f62012a6de177b240a/client/src/views/tx.js#L178-L220)

Since these strings require new translations, it should be handled in a separate task

Randy808:
Addressed regressions identified by @EddieHouston. Txs on the block page now show their confirmation status and total output amount in the footer:
Screenshot 2026-07-15 at 11 10 14 AM

client/src/views/tx.js
```diff
@@ -1,195 +1,414 @@
...
-        <div>{formatTime(tx.status.block_time)}</div>
-      </div>
-    ]}
+    <span>
+      <div className="transaction-table">
+        <div className="table-header">
+          <div className="table-header-icon-container">
+            <TxArrowsIcon />
+          </div>
+          <h1 className="table-header-title">Transaction Details</h1>
```
FedOken, line 252: Image

client/src/views/tx.js
```diff
@@ -1,195 +1,414 @@
...
+                      </div>
+                    </div>
+                  </div>
+                ) : null}
+              </div>
+            ) : null}
+
+            <div className="transaction-details-row">
+              <div className="tx-overview-panel">
+                <p className="tx-overview-panel-label">PRIVACY ANALYSIS</p>
```
FedOken, line 366: Image

Image
You might want to fix the arrows here

client/src/views/tx.js
```diff
@@ -1,195 +1,414 @@
...
-      <div>{`${formatNumber(tx.weight)} WU`}</div>
-    </div>
-    { tx.discount_vsize != null && <div>
-      <div>{t`Discount virtual size`}</div>
-      <div>{`${formatNumber(tx.discount_vsize)} vB`}</div>
-    </div> }
-    { tx.discount_weight != null && <div>
-      <div>{t`Discount weight units`}</div>
-      <div>{`${formatNumber(tx.discount_weight)} WU`}</div>
-    </div> }
```
FedOken, line 170: `discount_vsize` and `discount_weight`, I can't find them in the new version, they where removed?

client/src/components/block-grid.js
```diff
@@ -0,0 +1,87 @@
+const GRID_LENGTH = 15;
+const MAX_BLOCK_WEIGHT = 4_000_000;
```
FedOken:
```
export const getBlockPercentageUsed = blockWeight =>
  Math.round((blockWeight / 4000000) * 10000) / 100
```

We alredy have definition for 4000000. Either reuse what's already there or use the constant in both places

client/src/app.js
```diff
@@ -376,6 +377,10 @@ export default function main({ DOM, HTTP, route, storage, scanner: scan$, search
     // fetch single tx (including confirmation status)
     , goTx$.map(txid        => ({ category: 'tx',         method: 'GET', path: `/tx/${txid}` }))
 
+    // fetch the block containing a confirmed tx
+    , tx$.filter(tx         => tx && tx.status && tx.status.confirmed && tx.status.block_hash)
+        .map(tx             => ({ category: 'tx-block',   method: 'GET', path: `/block/${tx.status.block_hash}` }))
```
FedOken, line 382: If the request `/block/{hash}` returns an error (404, timeout, the block hasn't been indexed yet, etc.), the error is simply discarded.

Image

client/src/views/tx.js
Randy808: Yes, I addressed this here:
https://github.com/Blockstream/esplora/pull/638#issuecomment-4962350032

client/src/app.js
Randy808: I’m not sure I fully understand the point you’re making here. It sounds like an observation, but I’m not clear on what change you’re proposing. Could you clarify how this applies to the block request and how it differs from the request to the tx endpoint?

client/src/views/tx.js
Randy808: No text has been written for this so it was omitted.

Randy808: Left a comment asking for copy.

client/src/app.js
FedOken: Yes, of course. In the lines I highlighted, if we make an external call and that call fails, we’re not handling the error correctly. In other words, I tried entering an invalid URL for the external call in the code and saw an error on the front end. I think we need to handle this.
Previously, there was one call to retrieve data, and that was handled. Now a second one has been added, but we aren’t handling the second one.

Randy808: Both requests are handled in the same way. If either one fails, the user is redirected to the global error view. 

For example, if we changed the highlighted block request from:

`/block/${tx.status.block_hash}`

to:

`/block/abc/${tx.status.block_hash}`

we would get a result similar to the screenshot you shared. The same would happen if we changed the transaction request from:

`/tx/${txid}`

to:

`/tx/abc/${txid}`

Screenshot 2026-07-15 at 5 09 55 PM

Is there a different behavior you’d expect here, such as handling the block request failure without redirecting to the global error view? If so, I prefer consistent error handling over ad hoc error state designs.

Randy808: I added stylings to wrap the error message so it stays on the screen:
Screenshot 2026-07-15 at 11 25 54 PM

FedOken: lgtm

FedOken (approved):

takagimasao52-cmyk (approved):

PR #639, opened 2026-07-10
Documented the /block-template endpoint.

Randy808:
Added documentation for the `/block-template` endpoint added in https://github.com/Blockstream/electrs/pull/230


EddieHouston (approved):
👍

PR #642, opened 2026-07-20
Updated block page

Randy808:
```
# Bitcoin
export PORT=4999
npm run dev-server
```

```
# Liquid
export PORT=5000
npm run dev-server
```

----
Bitcoin:
Screenshot 2026-07-20 at 3 21 46 PM

Liquid:
Screenshot 2026-07-20 at 3 21 37 PM


client/src/views/block-details-card.js
```diff
@@ -46,8 +51,8 @@ const BlockDetailsCard = ({ className, block, confirmed }) => {
               )}
             </p>
```
FedOken, line 53: Here, it is better to adjust the design, font size, and capitalization.

Image

Image

client/src/views/block.js
```diff
@@ -1,180 +1,193 @@
...
+  tipHeight,
+  loading,
+  page,
+  txsStatus = makeStatus(b),
+  ...S
+}) =>
+  b &&
+  layout(
+    [
+      <div className="block-page-container">
```
FedOken, line 39: This is a global issue: when adapting to 1100px, the container breaks on all pages

Image

client/src/views/block-details-card.js
FedOken, line 90: Image

You can ask the designers, but in the meantime, here's what we can add ->

> Block fill is the percentage of a Bitcoin block's maximum capacity (~4M weight units) that's actually used by transactions — high fill means high demand and rising fees, low fill means spare capacity and cheap fees.

client/src/views/block.js
```diff
@@ -1,180 +1,193 @@
...
-                <div>{t`Bits`}</div>
-                <div className="mono">{formatHex(b.bits)}</div>
-              </div>
-            , <div>
-                <div>{t`Difficulty`}</div>
-                <div className="mono">{formatNumber(b.difficulty)}</div>
-              </div>
-            , <div>
-                <div>{t`Nonce`}</div>
-                <div className="mono">{formatHex(b.nonce)}</div>
```
FedOken, line 102: Shouldn't we keep these metrics?

client/src/views/block.js
```diff
@@ -1,180 +1,193 @@
...
-      {btnDetails(b.id, openBlock == b.id, page.query, t)}
-
-      <div className="stats-table font-p2">
-        <div>
-          <div>{t`Height`}</div>
-          <div><a href={`block/${b.id}`}>{b.height}</a></div>
-        </div>
-        <div>
-          <div>{t`Status`}</div>
-          <div>{!status ? '' : !status.in_best_chain ? t`Orphaned` : tipHeight ? t`In best chain (${tipHeight-b.height+1} confirmations)` : t`In best chain`}</div>
```
FedOken, line 59: It's strange that the number of confirmations has disappeared, in theory, that's important for a network like Bitcoin.

client/src/views/block.js
```diff
@@ -1,180 +1,193 @@
...
-        { openBlock == b.id && [
+        <div className="block-box">
+          <div className="detail-grid">
+            <div className="detail-grid-row">
+              <div className="detail-field">
+                <p className="detail-field-label">VERSION</p>
+                <div className="detail-field-content">
+                  <p>{formatHex(b.version)}</p>
+                </div>
+              </div>
```
FedOken, line 89: It appears many times, it may be a component.
```
<div className="detail-field">
  <p className="detail-field-label">LABEL</p>
  <div className="detail-field-content">
    {content}
  </div>
</div>
```

client/src/views/block.js
Randy808: A case can be made that block depth is not necessary when confirmations for individual transactions are shown. As stated in our call, this is an issue for you to take up with design.

client/src/views/block.js
Randy808: I am okay with the change, but this is another issue for you to take up with design.

client/src/views/block-details-card.js
Randy808: This has an analogue message:
https://www.figma.com/design/shpGaiWldAQ0JvBnS24ABY?node-id=1770-3759&m=dev#1803015996

I'll just remove the word 'next'. The design seems to want to keep tooltips short.

client/src/views/block-details-card.js
Randy808: Done

client/src/views/block.js
Randy808: Changed breakpoint to 1328px

client/src/views/block-details-card.js
FedOken: I don't get it, you don't have that hint

image

Randy808: I never said I did nor did I resolve this issue.

Randy808: I just pushed the change with the tooltip.

client/src/views/block.js
Randy808: I'll defer making this into a component and include it in the follow-up PR (currently in draft [here](https://github.com/Blockstream/esplora/pull/643)). The change in that one is smaller making the added-component easier to review in context.

FedOken (approved):

takagimasao52-cmyk (approved):

PR #643, opened 2026-07-21
Address page

Randy808:
```
# Bitcoin
export PORT=4999
npm run dev-server
```

```
# Liquid
export PORT=5000
npm run dev-server
```

----
Bitcoin:
Screenshot 2026-07-21 at 3 11 39 PM

Liquid:
Screenshot 2026-07-21 at 3 11 59 PM


client/src/views/addr.js
```diff
@@ -1,161 +1,276 @@
...
+      : addr.display_addr,
+    show_qr = !process.env.IS_ELEMENTS || is_confidential;
 
   return layout(
     [
       <div className="addr-page">
-        <div className="container">
-          <div className="addr-header-layout">
-            <div className="addr-header-main">
-              <h1 className="font-h2">{t`Address`}</h1>
```
FedOken, line 45: Image

1. We need to add a tooltip with a question mark.

2. This is for the designers, but it’s worth asking about, since it looks like a link but isn’t one, it’s very strange.

3. I think we should trim the zeros.

4. Will the text be formatted with a comma as the separator? 2472 -> 2,472

FedOken: There are no other questions.

Randy808: I can address (2) and (4). I've requested copy for (1) from design and copy-pasted your comment on (2) to Figma for design

FedOken: Points (1) and (2) are clear, this is a question for the designers, but it's not a big deal. What about points (3) and (4)?
About (4) it's already in the design. If we're going to add something like this, then in theory it should apply to all numerical values. We can leave this in another PR.

Randy808: For (1), **there is no tooltip \*_message_\* for the address.** I left a message asking for copy [on Figma and cc'ed you](https://www.figma.com/design/shpGaiWldAQ0JvBnS24ABY?node-id=2965-6533&m=dev#1860023245)

Randy808: Also I meant I can address (3)* and (4). Putting '(2)' was a typo.

philippem:
on FF the "SEGWIT FEE SAVINGS" breaks "SegWit-Bech32" across a line
image

client/src/views/addr.js
Randy808: I addressed all of them except trimming the 0s. I think if we remove trailing 0s it should be its own deliberate change because, on second thought, I'm unsure if it would be a good decision to.

Randy808:
@philippem [Removed](https://github.com/Blockstream/esplora/pull/643/changes#diff-3f710416dbe04f0fa44aeaef9817a13d98f8ed47f63c3c1615adfdddcca8ec06L4196) the line break rule that allowed it to happen anywhere

agoodminute:
`/address/39C7fxSzEACPjM78Z7xdPxhf7mKxJwvfMJ`

1. `cursor: pointer;` it should also be removed; I think it's confusing, especially next to the copy icon.
2. If the numbers are large, the QR code is shifted downward for some addresses.

image

client/src/views/addr.js
FedOken: Okay, we can discuss this separately.

philippem:
Let's address the layout for the long address description in a later PR.

philippem (approved):

takagimasao52-cmyk (approved):

PR #644, opened 2026-07-23
Add drivechain peg activity explorer

ekulkisnek:
## Summary

- add an Elements-specific peg activity view for deposits, withdrawals, bundle commitments, and acknowledgements
- expose the view through navigation and document the corresponding API contract
- lock the production frontend dependencies and add CI coverage for reproducible builds

## Why

LayerTwoLabs Liquid Signet exposes BIP300 peg lifecycle data that was not visible in the standard Esplora interface. This gives operators a read-only explorer view of that lifecycle while retaining the existing Esplora transaction and asset views.

## Validation

- production frontend build completed successfully
- live UI served on localhost and returned the expected explorer application
- Esplora API and Elements RPC agreed at height 2673 and on the exact tip hash
- responsive peg activity UI was verified against the indexed Liquid Signet fixture

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,42 @@
+name: Build
+
+on:
+  push:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  locked-production-build:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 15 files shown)


ekulkisnek:
Closing this upstream PR. The Liquid drivechain adaptation will be maintained as a separate repository instead.

takagimasao52-cmyk (approved):

PR #653, opened 2026-08-12
Added fee market components.

Randy808:
```
# Bitcoin
export PORT=4999
npm run dev-server
```

```
# Liquid
export PORT=5000
npm run dev-server
```

---

Bitcoin:
fee btc | Screenshot 2026-08-13 at 11 58 37 AM

Liquid:
fee lqd | Screenshot 2026-08-13 at 11 58 06 AM

### DIFF
--- a/client/src/lib/fees.js
+++ b/client/src/lib/fees.js
@@ -1,11 +1,22 @@
+import {
+  feeEstimateTargets,
+  satoshisPerBitcoin,
+  typicalDiscountedConfidentialTransactionVsize,
+  typicalNativeSegwitTransactionVsize,
+} from '../const'
+
 const MAX_BLOCK_VSIZE = 1000000
+    , typicalTransactionVsize = process.env.IS_ELEMENTS
+        ? typicalDiscountedConfidentialTransactionVsize
+        : typicalNativeSegwitTransactionVsize
 
 export const getFeeTierBoundaries = feeEst => {
-  const low = feeEst && feeEst[12]
-    , high = feeEst && feeEst[3]
+  const lowMaximum = feeEst && feeEst[feeEstimateTargets.low]
+    , mediumMaximum = feeEst && feeEst[feeEstimateTargets.average]
 
-  return Number.isFinite(low) && Number.isFinite(high) && low >= 0 && high >= 0
-    ? { low, high: Math.max(low, high) }
+  return Number.isFinite(lowMaximum) && Number.isFinite(mediumMaximum)
+      && lowMaximum >= 0 && mediumMaximum >= 0
+    ? { lowMaximum, mediumMaximum: Math.max(lowMaximum, mediumMaximum) }
     : null
 }
 
@@ -13,9 +24,9 @@ export const feeRateClass = (feerate, feeEst) => {
   const boundaries = getFeeTierBoundaries(feeEst)
   if (!Number.isFinite(feerate) || !boundaries) return ""
 
-  return feerate <= boundaries.low
+  return feerate <= boundaries.lowMaximum
     ? "success"
-    : feerate <= boundaries.high
+    : feerate <= boundaries.mediumMaximum
       ? "warning"
       : "danger"
 }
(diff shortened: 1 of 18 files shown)


client/src/lib/fees.js
```diff
@@ -1,8 +1,14 @@
...
+  satoshisPerBitcoin,
+} from '../const'
+
 const MAX_BLOCK_VSIZE = 1000000
 
 export const getFeeTierBoundaries = feeEst => {
-  const low = feeEst && feeEst[12]
-    , high = feeEst && feeEst[3]
+  const low = feeEst && feeEst[feeEstimateTargets.low]
+    , high = feeEst && feeEst[feeEstimateTargets.average]
```
agoodminute: You take `feeEstimateTargets.average` but marked it as `hight`. This could lead to misunderstandings in feature.
```
const feeEstimateTargets: {
    low: number;
    average: number;
    high: number;
}
```

client/src/lib/market.js
```diff
@@ -0,0 +1,9 @@
+export const getBitcoinPrices = marketChart =>
+  ((marketChart && marketChart.prices) || [])
+    .map(price => price && price[1])
+    .filter(Number.isFinite)
```
agoodminute, line 4: Better to add check for array value, and we can simplify code like `marketChart && marketChart.prices` to `marketChart?.prices`
```
export const getBitcoinPrices = marketChart =>
  (Array.isArray(marketChart?.prices) ? marketChart.prices : [])
    .map(price => price && price[1])
    .filter(Number.isFinite)
```

client/src/views/fee-market.js
```diff
@@ -0,0 +1,53 @@
...
+            feerate,
+          );
+
+          return (
+            <InfoCard
+              className={`fee-market-${key}`}
+              title={title}
+              tooltip={tooltip}
+              value={formatFeeRate(feerate, unavailable)}
+              footer={formatUsd(feeUsd, unavailable)}
```
agoodminute, line 46: We can use default fallback instead of `unavailable`. The same in other places

client/src/lib/fees.js
```diff
@@ -38,6 +44,16 @@ export function getConfEstimate(fee_estimates, feerate) {
   return target_est ? target_est[0] : -1
 }
 
+// Estimate the USD cost of a typical native SegWit transaction at a given fee-rate.
+export function estimateNativeSegwitTransactionFeeUsd(bitcoinPrice, feerate) {
+  return Number.isFinite(bitcoinPrice) && bitcoinPrice >= 0
+      && Number.isFinite(feerate) && feerate >= 0
+    ? (bitcoinPrice / satoshisPerBitcoin) *
+        feerate *
+        averageNativeSegwitTransactionVsize
```
agoodminute: We render this block for both networks.

On the Liquid Network, the concept of a "Native SegWit transaction" does not apply in the same way as on Bitcoin. Liquid uses a modified transaction architecture built on the Elements codebase, where default transfers are Confidential Transactions that bundle cryptographic range proofs and Pedersen commitments for privacy.

A standard single-input, dual-output Liquid transaction is significantly larger than a plain Bitcoin Native SegWit (P2WPKH) transaction - averaging roughly 1500 to 2000 bytes/vBytes instead of ~141 vBytes due to the mandatory inclusion of size-heavy range proofs (bulletproofs) that hide asset types and amounts.

Randy808: As of [ELIP200](https://blog.liquid.net/confidentiality-made-affordable-elip-200-unlocks-up-to-90-lower-fees-on-liquid/) the surjection + range proof (still borromean ring sigs as we haven't changed consensus to accommodate bulletproofs yet), the nonce commitment, and amount commitment have been discounted to make a standard transaction's weight closer to bitcoin's. I'll double check a standard transaction's calculation after the discount, but it should be very close to bitcoin

client/src/lib/fees.js
Randy808: The boundaries don't quite represent the exact value calculations. 

For example:
- if a value is below the low boundary, it's low
- if a value is greater than low but less than average, it's avg
- if a value is higher than avg, it's high

We don't need the high value to classify individual transactions. We only need a reasonably high upper-bound estimate to show to the user. But I see your point that it may be confusing. I'll see if I can add a comment about it.

philippem: it does show up as high=average in the UI which may be confusing/misleading

Image

Randy808: Yeah, the estimates between getting into the next block (high fee) vs the next 3 (avg fee) seem to track pretty closely
Screenshot 2026-08-14 at 11 07 59 AM

client/src/views/fee-market.js
Randy808: Looking at it closer, `unavailable` is used for future localization purposes

client/src/lib/fees.js
Randy808: In addition to the dummy bytes included in the discount calculation, I forgot about the 33 bytes reserved for the asset id. The template tx size value should be updated now

client/src/lib/fees.js
Randy808: The variables were given better names to represent what they are (`lowMaximum` and `mediumMaximum`)

philippem (approved):

PR #654, opened 2026-08-12
Added 'High-Value Assets' component.

Randy808:
```
# Bitcoin
export PORT=4999
npm run dev-server
```

```
# Liquid
export PORT=5000
npm run dev-server
```

Liquid:
hva | Screenshot 2026-08-13 at 11 57 09 AM


client/src/components/high-value-assets.js
```diff
@@ -0,0 +1,53 @@
+import { CurrencyDollarIcon } from "./icons";
```
agoodminute, line 1: 1. When you hover over it, the highlight on the asset is smaller than the block itself. We can try limiting it to one decimal place.
2. Should it be sorted in descending order?
Image

client/src/components/high-value-assets.js
```diff
@@ -0,0 +1,53 @@
...
+import { highValueAssetDefinitions as assets } from "../const";
+import {
+  calculateIssuedDollarAmount,
+  formatDollarAmount,
+} from "../lib/high-value-assets";
+
+const staticRoot = process.env.STATIC_ROOT || "";
```
agoodminute: This isn't the first time this has happened, but as part of the future refactoring, I think we should add all the variables that access `process.env...` to the const.js file and import their values from this file

Randy808: Moved into const file

client/src/components/high-value-assets.js
Randy808: Changed to show one decimal and sorted by value

philippem: maybe we shouldn't show any decimals, and round to whole number these numbers are volatile, already big.

philippem (dismissed):

agoodminute (dismissed):

client/src/components/high-value-assets.js
zhennangaomu00-blip: どうすれば良いのか押して下さい

philippem (approved):

PR #656, opened 2026-08-15
Standardize polling

Randy808:
To alleviate the burden on user rate limits, I reduced the polling cadence for some endpoints to something more manageable (and standardized polling intervals in const.js).

Here's a summary of the changes:

- Standardized global chain-tip polling at 30s, while retaining navigation-triggered refreshes on block, transaction, and address pages, rate-limited to at most
    once every 5s
- Changed recent mempool transaction polling from 5s to 30s
- Changed mempool summary polling from 30s to 60s on mempool and unconfirmed-transaction pages
- Changed fee-estimate polling to 60s on mempool, recent-transactions, and unconfirmed-transaction pages
- Changed block-list refreshes to be tip-triggered, with a 10-minute fallback for same-height reorgs
- Replaced timed confirmed-peg polling with block-triggered requests; retained 60s polling for pending peg transactions
- Changed market chart refreshes from 60s to 10 minutes
- Changed high-value asset refreshes from 60s to 10 minutes

---

```
# Bitcoin
export PORT=4999
npm run dev-server
```

```
# Liquid
export PORT=5000
npm run dev-server
```

### DIFF
--- a/client/src/app.js
+++ b/client/src/app.js
@@ -13,6 +13,7 @@ import {
   blockTxsPerPage,
   blocksPerPage,
   difficultyPeriod,
+  pollIntervalsMs,
   showHighValueAssets,
   showPegData,
   blockGridTransactionSelectEvent
@@ -30,7 +31,8 @@ import {
     isHash256,
     makeAddressQR,
     tickWhileFocused,
-    tickWhileViewing,
+    pollWhileActive,
+    pollWhileViewing,
     updateBlocks,
     calculateFeerates,
     calculateOverpayment,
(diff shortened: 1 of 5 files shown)


client/src/app.js
```diff
@@ -137,6 +140,30 @@ export const scheduleDashboardBlockTemplatePolls = (
   .filter(({ view }) => view == 'dashBoard')
   .map(({ pollIndex }) => pollIndex)
 
+export const dashboardNewBlocks = (page$, latestBlock$) =>
+  page$
+    .switchMap(page => page && page.pathname == '/'
```
agoodminute, line 145: Should we use `view$ == 'dashBoard'` instead of `page.pathname == '/'`?

client/src/app.js
```diff
@@ -42,7 +44,8 @@ const highValueAssetCategory = assetId => `dashboard-high-value-asset-${assetId}
     , highValueAssetPriceCategory = assetId => `dashboard-high-value-asset-price-${assetId}`
     , apiBase = (process.env.API_URL || '/api').replace(/\/+$/, '')
     , bitcoinMarketChartUrl = process.env.BITCOIN_MARKET_CHART_URL || 'https://api.coingecko.com/api/v3/coins/bitcoin/market_chart?vs_currency=usd&days=1&interval=hourly'
-    , blockTemplatePollIntervalMs = 30000
+    , blockTemplatePollIntervalMs = pollIntervalsMs.fast
+    , tipRequestThrottleMs = 5000
```
agoodminute, line 48: It wasn't added to the new `pollIntervalsMs` table, even though the whole point of the PR was to consolidate all the intervals in one place. It's a minor issue.

Randy808: The throttle is different from the polling. It represents how often the request can be made (mainly to limit event-triggered requests like navigating to tx page)

client/src/app.js
Randy808: `view$` has a loading state that would delay the establishment of the stream. Delaying the establishment of the stream might cause us to skip a new block ('new' being any block after the initial one) instead of just the initial one.

agoodminute (approved):

PR #684, opened 2026-09-20
docs: add missing run subcommand to Tor example

agammann:
The Tor container example starts with `docker -d`, which Docker rejects with `unknown shorthand flag: 'd' in -d`. Add the missing `run` subcommand so the existing detach, name, image and Tor arguments are passed to the container command.

Validation: reproduced the original CLI error and confirmed that `docker run -d --name hidden_service --help` accepts the options. `git diff --check` passes. No image was pulled or container started.

CI note: the GitHub-to-GitLab trigger failed with HTTP 404 before any checkout or project tests ran ([job log](https://github.com/Blockstream/esplora/actions/runs/35520749034/job/106104285057)). The local Docker command validation above passed.


agammann:
I checked the failed [trigger-gitlab job](https://github.com/Blockstream/esplora/actions/runs/35520749034/job/106104285057). It reaches the GitLab pipeline trigger for this PR's commit, but the request returns HTTP 404 and curl exits with code 22. The job stops before it can report a GitLab test result.

The log does not establish the cause of the 404. Could a maintainer check the GitLab trigger configuration/access and rerun the job when appropriate? I have not changed the workflow as part of this PR.

